PR #55, opened 2019-11-22
maven github action

jmcrfp:
## Status
**READY**

## Migrations
NO

## Description
Simple maven github action

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,17 @@
+name: Java CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml
```
mcpierce: Can you please add "-Pci" as a flag to the command line so Maven uses the continuous integration profile?

mcpierce (changes requested):

.github/workflows/maven.yml
jmcrfp: done

mcpierce (approved):

Issue #57, opened 2019-11-22
Importing simple lists from comicrack


jmcrfp:
As a user I want simple lists (manual created ones, not the smart lists) to be imported from the comicrack database into comixed

jmcrfp:
Implemented

PR #65, opened 2019-11-24
[Issue #57] - Imports manual lists from comicrack database into comixed

jmcrfp:
## Status
**READY**

## Migrations
NO

## Description
Imports manual lists from comicrack database into comixed.

### DIFF
--- a/comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
+++ b/comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
@@ -29,6 +29,7 @@ import org.comixed.importer.adaptors.ImportAdaptorException;
 import org.comixed.model.library.Comic;
 import org.comixed.model.user.ComiXedUser;
 import org.comixed.repositories.ComiXedUserRepository;
+import org.comixed.service.library.ComicException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


mcpierce (dismissed):
There are merge conflicts. Can you fix them so I can do a review and merge, please?

comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
```diff
@@ -74,7 +77,7 @@ public void setImportUser(String user) {
    *
    * @throws ProcessorException if a processing error occurs
    */
-  public void process(String source) throws ProcessorException {
+  public void process(String source) throws ProcessorException, ComicException {
```
mcpierce: I don't think we should add the ComicException to the method signature. Instead, can we have the implementation catch the ComicException and rethrow it as a ProcessorException? That way we know each step of the way in the stacktrace where something bad happened.

mcpierce (changes requested):
Just one change requested, all else looks good.

comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
jmcrfp: You're right. Changed it

mcpierce (approved):

PR #80, opened 2019-12-01
Fix compilation issues in windows wsl and some cleanups

jmcrfp:
## Status
**READY**

## Migrations
NO

## Description
Fixes somes issues that would lock up build in windows wsl, so upgraded the plugins
Also removed some warnings
Also removed some unnecessary pom tags that were showing up on intellij


mcpierce (approved):
Tested the build on both Windows and Linux to ensure all platforms were good, and the builds passed.

PR #90, opened 2019-12-07
Cache image files outside of the comic archive

mcpierce:
## Status
**READY**

## Migrations
YES

## Description
Cache image files outside of the comic archive after initial access to make them easier to load afterward.


jmcrfp:
The testFindByHash is failing

jmcrfp (changes requested):
The test testFindByHash is failing

mcpierce:
> The test testFindByHash is failing

Yeah, forgot to add the test image to the commit. That's fixed now.

comixed-services/src/main/java/org/comixed/service/library/PageCacheService.java
```diff
@@ -0,0 +1,79 @@
...
+    this.logger.debug("Searching for cached image: hash={}", hash);
+
+    final File file = this.getFileForHash(hash);
+    if (file.exists() && !file.isDirectory()) {
+      this.logger.debug("Loading cached image content: {} bytes", file.length());
+      try {
+        return IOUtils.readFully(new FileInputStream(file), (int) file.length());
+      } catch (Exception error) {
+        this.logger.error("Failed to load cached image", error);
+      }
```
jmcrfp, line 52: Add a finally close of the of the open file so that in case of exception we don't get leaks

mcpierce: Not as a finally, which require another level of try/catch coding. But I refactored it to hold a reference to the input stream so it can be explicitly closed.

jmcrfp (approved):

Issue #270, opened 2020-05-15
Using invalid credentials to log in returns a "Successfully logged out" toaster message.


bareheiny:
As per title.

I attempted to log into CX using an invalid user / password combination, and noted the "Successfully logged out" message was displayed, rather than the expected red (unsure of exact message) toaster.

BRUCELLA2:
I took a quick look at this bug.
It seems that when the username/password combination is invalid, the back returns a 401 error.
The 401 error is processed in hrx.interceptor and calls the startLogout method of the authenticationAdaptor.

I think we can call a new method (authenticationFailed) when error 401 is returned. This method can then dispatch the AuthLoginFailed action and we need to add the @Effect for authentication that fails.

mcpierce:
@BRUCELLA2 I think you're spot on with that. Would you be able to provide a PR to do that?

BRUCELLA2:
@mcpierce i think i can provide a PR, yes.

mcpierce:
@bareheiny A fix was provided by @BRUCELLA2 -- can you verify on the next build that this bug is fixed and close the ticket? Thanks.

bareheiny:
Looks good, closing.

PR #271, opened 2020-05-22
[Issue #270] Fix wrong message when using invalid credentials to log.

BRUCELLA2:
## Status
READY

## Migrations
NO

## Description
Fix wrong message when using invalid credentials to log.

### DIFF
--- a/comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
+++ b/comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
@@ -206,6 +206,17 @@ describe('AuthenticationAdaptor', () => {
     });
   });
 
+  describe('when login failed', () => {
+    beforeEach(() => {
+      spyOn(store, 'dispatch');
+      auth_adaptor.authenticationFailed();
+    });
+
+    it('fires an action', () => {
+      expect(store.dispatch).toHaveBeenCalledWith(new AuthActions.AuthLoginFailed());
+    });
+  });
+
   describe('when setting a user preference', () => {
     beforeEach(() => {
       spyOn(store, 'dispatch');
--- a/comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
+++ b/comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
@@ -38,7 +38,7 @@ describe('AuthenticationAdaptor', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [StoreModule.forRoot({ auth_state: reducer })],
-      providers: [AuthenticationAdaptor]
+      providers: [AuthenticationAdaptor],
     });
 
     auth_adaptor = TestBed.get(AuthenticationAdaptor);
@@ -113,6 +113,18 @@ describe('AuthenticationAdaptor', () => {
       expect(store.dispatch).toHaveBeenCalledWith(
         new AuthActions.AuthCheckState()
       );
+    });
+  });
+
+  describe('when sending the loading data', () => {
+    beforeEach(() => {
+      store.dispatch(
+        new AuthActions.AuthSubmitLogin({ email: EMAIL, password: PASSWORD })
+      );
+    });
+
+    it('returns true for the authenticating', () => {
+      expect(auth_adaptor.authenticating).toBeTruthy();
     });
   });
 
@@ -213,7 +225,9 @@ describe('AuthenticationAdaptor', () => {
     });
 
     it('fires an action', () => {
-      expect(store.dispatch).toHaveBeenCalledWith(new AuthActions.AuthLoginFailed());
+      expect(store.dispatch).toHaveBeenCalledWith(
+        new AuthActions.AuthLoginFailed()
+      );
     });
   });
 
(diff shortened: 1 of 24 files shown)


comixed-frontend/src/app/user/adaptors/authentication.adaptor.ts
```diff
@@ -22,7 +22,7 @@ import { Roles } from 'app/models/ui/roles';
 import { AppState, User } from 'app/user';
 import {
   AuthCheckState,
-  AuthHideLogin,
+  AuthHideLogin, AuthLoginFailed,
```
mcpierce: Could you run Prettier on this commit and refresh the PR?

comixed-frontend/src/app/user/adaptors/authentication.adaptor.ts
```diff
@@ -137,6 +137,11 @@ export class AuthenticationAdaptor {
     this.store.dispatch(new AuthLogout());
   }
 
+  authenticationFailed(): void {
+    this.store.dispatch(new AuthLoginFailed());
+    this.store.dispatch(new AuthHideLogin());
```
mcpierce: I don't think, in the case of a failed login attempt, we should hide the login dialog. We should only hide the login dialog if the user explicitly clicks **Cancel** on it.

comixed-frontend/src/app/xhr.interceptor.ts
```diff
@@ -77,7 +77,7 @@ export class XhrInterceptor implements HttpInterceptor {
             return;
           case 401:
             this.logger.error('[XHR] user not authenticated:', error);
-            this.authenticationAdaptor.startLogout();
+            this.authenticationAdaptor.authenticationFailed();
```
mcpierce: Do you think we could add a boolean field to AuthenticationState, called "authenticating", to indicate whether or not we're in the process of doing a login?

Then here the code could do something like:
```
if(this.authenticationAdaptor.authenticating) {
  this.logger.debug('Authentication attempt faied');
  this.authenticationAdaptor.authenticationFailed();
} else {
  this.logger.debug('No valid auth token');
  this.authenticationAdaptor.startLogout();
}
```

I think this would treat the two scenarios (trying to login vs. trying to access a page with an expired or no authentication token) differently and let us show the right toaster message.

mcpierce (changes requested):
All in all looks good. A few changes are requested then we can merge this.

And please be sure to add yourself to the CONTRIBUTORS.md file as a contributor the project.

comixed-frontend/src/app/user/adaptors/authentication.adaptor.ts
BRUCELLA2: I've added the Prettier plugin in IntelliJ but I'm not sure if the configuration is correct because there are some modifications on code I didn't touch.

comixed-frontend/src/app/user/adaptors/authentication.adaptor.ts
BRUCELLA2: Ok.

comixed-frontend/src/app/xhr.interceptor.ts
BRUCELLA2: Sounds like a good idea indeed.

comixed-frontend/src/app/user/adaptors/authentication.adaptor.ts
mcpierce: Prettier should pickup the configuration in the package.json file. We just need to keep the code formatted so commits are solely functional changes.

Be sure to only run the plugin on Typescript code, though. It gets weird on HTML (I haven'tgotten  that configured yet).

comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
```diff
@@ -38,7 +38,7 @@ describe('AuthenticationAdaptor', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [StoreModule.forRoot({ auth_state: reducer })],
-      providers: [AuthenticationAdaptor]
+      providers: [AuthenticationAdaptor],
```
mcpierce: Still needs prettier run to clean up a few trailing commas like above.

(sorry if this feels like nitpicking, but I'm trying to keep things well-formatted)

To make it easier (and don't worry if it touches other files you didn't change, since I probably missed formatting a few files) you can do the following from the comixed-frontend directory:

```
prettier --write 'src/app/**/*.ts'
```

and commit all of the changes.

mcpierce (changes requested):
Just need to do that final run of Prettier on the code, and add yourself to the CONTRIBUTORS.md file so you get credit for your work, and we can merge this.

comixed-frontend/src/app/user/adaptors/authentication.adaptor.spec.ts
BRUCELLA2: It's ok I hope :) I like when the things are well-formatted too so it's not a problem.
I understand now why my prettier clean up was not good before. I used the latest version 2.0.5 and there are changes in the default behavior especially on trailingComma, arrowParens, and endOfLine options. (https://prettier.io/blog/2020/03/21/2.0.0.html)

mcpierce (approved):

PR #355, opened 2020-06-15
SonarCloud issue fixes

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Fixed a few issues reported by SonarCloud.

### DIFF
--- a/comixed-library/src/main/java/org/comixed/adaptors/archive/SevenZipArchiveAdaptor.java
+++ b/comixed-library/src/main/java/org/comixed/adaptors/archive/SevenZipArchiveAdaptor.java
@@ -69,8 +69,11 @@ public class SevenZipArchiveAdaptor extends AbstractArchiveAdaptor<SevenZFile> {
     archive.write(content);
     archive.closeArchiveEntry();
 
-    log.debug("Cleaning up the temporary file");
-    tempFile.delete();
+    if (tempFile.delete()) {
+      log.debug("Cleaning up the temporary file");
+    } else {
+      log.debug("Failed to delete temporary file: {}", tempFile.getAbsolutePath());
+    }
   }
 
   @Override
--- a/comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
+++ b/comixed-importer/src/main/java/org/comixed/importer/ImportFileProcessor.java
@@ -55,14 +55,14 @@ public class ImportFileProcessor {
   @Autowired private ComiXedUserRepository userRepository;
 
   public void setReplacements(List<String> replacements) {
-    this.log.debug("Processing {} replacement rules", replacements.size());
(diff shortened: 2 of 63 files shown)


comixed-library/src/main/java/org/comixed/adaptors/archive/SevenZipArchiveAdaptor.java
```diff
@@ -69,8 +69,11 @@ private void addFileToArchive(SevenZOutputFile archive, String filename, byte[]
     archive.write(content);
     archive.closeArchiveEntry();
 
-    this.log.debug("Cleaning up the temporary file");
-    tempFile.delete();
+    if (tempFile.delete()) {
+      this.log.debug("Cleaning up the temporary file");
```
BRUCELLA2: lombok generate this logger : 
private static final Logger log = org.apache.logging.log4j.LogManager.getLogger(SevenZipArchiveAdaptor.class);
so I think we can access to the log without the "this".

comixed-library/src/main/java/org/comixed/adaptors/archive/SevenZipArchiveAdaptor.java
```diff
@@ -69,8 +69,11 @@ private void addFileToArchive(SevenZOutputFile archive, String filename, byte[]
...
+    } else {
+      this.log.debug("Failed to delete temporary file: {}", tempFile.getAbsolutePath());
```
BRUCELLA2: lombok generate this logger : 
private static final Logger log = org.apache.logging.log4j.LogManager.getLogger(SevenZipArchiveAdaptor.class);
so I think we can access to the log without the "this".

BRUCELLA2 (changes requested):
Quality Gate failed, we are < 80% of coverage of New Code. However, in view of what is not covered I think 76% coverage is acceptable. Do you agree with that?

comixed-library/src/main/java/org/comixed/adaptors/archive/SevenZipArchiveAdaptor.java
mcpierce: Could you do a global search-and-replace to replace "this.log" in the Java code with "log" and submit a PR?

BRUCELLA2: Yes i can.

mcpierce:
@BRUCELLA2 I've added some additional changes to hopefully wrap up the SonarCloud issues.

BRUCELLA2 (approved):
It's ok for me. I approve this PR

BRUCELLA2 (approved):

BRUCELLA2 (approved):

Issue #376, opened 2020-06-28
Prompt the user for confirmation before logging out


mcpierce:
When the user selects the logout button, the system should prompt to confirm they really want to log out.

bareheiny:
Is it all feasible to allow users to disable confirmation prompts (maybe for everything except irreversible actions such as deleting comics etc.)?

Personally, I'll find myself more annoyed by a "are you sure" prompt than an accidental log-out (or comic scrape commit).

mcpierce:
It's possible. We'd have to add it as a separate feature request, but it's not out of the question.

PR #377, opened 2020-06-29
Added prompt the user for confirmation before logging out (#376)

BRUCELLA2:
## Status
**READY**

## Migrations
NO

## Description
Added prompt the user for confirmation before logging out

### DIFF
--- a/comixed-frontend/src/assets/i18n/en/app.json
+++ b/comixed-frontend/src/assets/i18n/en/app.json
@@ -90,8 +90,6 @@
       "no-description": "No description available...",
       "delete-comic-title": "Delete This Comic?",
       "delete-comic-question": "Are you sure? This will not delete the file, but only remove it from the library.",
-      "confirm-logout-title": "Logout",
-      "confirm-logout": "Are you sure you want to log out?",
       "rescan-library-title": "Rescan Library?",
       "rescan-library-message": "Are you sure? This could take a very long time to complete..."
     },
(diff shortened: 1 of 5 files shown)


comixed-frontend/src/assets/i18n/en/app.json
```diff
@@ -90,6 +90,8 @@
       "no-description": "No description available...",
       "delete-comic-title": "Delete This Comic?",
       "delete-comic-question": "Are you sure? This will not delete the file, but only remove it from the library.",
+      "confirm-logout-title": "Logout",
```
mcpierce: I'm not sure we would want to put the logout confirmation message into the library i18n namespace. Can we instead move this to the "global" namespace under a new collection called "confirmations"? Then set the two parts as "header" and "message" entries in there, which is in line with how other confirmations are translated:

```
  "global": {
    "errors": {
      "form-field-required": "{name} is required."
    },
    "tooltip": {
      "unprocessed-comic": "This comic is scheduled for processing.",
      "deleted-comic": "This comic is marked for deletion."
    },
   "confirmation": {
        "logout": {
            "header": "Logout",
            "message": "Are you sure you want to log out?"
        }
    }
  },
```

mcpierce (changes requested):
All looks good. Just need to move the i18n entries and we can merge it.

comixed-frontend/src/assets/i18n/en/app.json
BRUCELLA2: Modification done

Issue #378, opened 2020-06-29
Admins can specify the filename and directory name rules when moving comics.


mcpierce:
This issue depends on #21.

When consolidating the library, the admin can provide a rule that will be used when creating the directories and when renaming the comic files when they're moved. The rule will use the following values:

Variable | Replaced With
---------|--------------
$PUBLISHER | The comic's publisher
$SERIES | The comic's series
$VOLUME | The comic's volume
$YEAR | The comic's cover year
$COVERDATE | The comic's cover date (in the format "YYYY-MM-DD")
$ISSUE | The comic's issue number

mcpierce:
There are instances where the old renaming rules are still being applied.

Issue #365, opened 2020-06-25
Review the REST APIs and apply any security requirements to them


mcpierce:
It would be a great improvement to add user role checks to APIs to ensure that only users with the appropriate authorization can access them; i.e., a user with the READER role should not be able to access admin features, etc.

Additionally, to enforce these checks, we should have unit tests for the REST APIs that use [REST Assured](http://rest-assured.io/) to validate them.

PR #383, opened 2020-06-30
Users can move their entire library to a new directory

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Added a new REST API for starting the moving process. Added a worker task to create individual move tasks for each comic in the library. Updated the frontend to take as input the new root directory and renaming rules, but the latter is not currently used (that's going to be in #378), so the comics are moved to a directory and filename of $ROOT/PUBLISHER/SERIES/VOLUME/SERIES vVOLUME #ISSUE.cb?

### DIFF
--- a/comixed-frontend/src/app/library/adaptors/library.adaptor.spec.ts
+++ b/comixed-frontend/src/app/library/adaptors/library.adaptor.spec.ts
@@ -47,13 +47,13 @@ import {
   LibraryClearImageCache,
   LibraryClearImageCacheFailed,
   LibraryComicsConverting,
-  LibraryConsolidate,
-  LibraryConsolidated,
-  LibraryConsolidateFailed,
+  LibraryComicsMoved,
   LibraryConvertComics,
   LibraryConvertComicsFailed,
   LibraryGetUpdates,
   LibraryImageCacheCleared,
+  LibraryMoveComics,
+  LibraryMoveComicsFailed,
   LibraryUpdatesReceived
 } from '../actions/library.actions';
 import { LibraryAdaptor } from './library.adaptor';
@@ -76,6 +76,9 @@ describe('LibraryAdaptor', () => {
   const ARCHIVE_TYPE = 'CBZ';
   const RENAME_PAGES = true;
   const READING_LISTS = [READING_LIST_1, READING_LIST_2];
+  const DIRECTORY = '/Users/comixedreader/Documents/comics';
+  const RENAMING_RULE =
+    '$PUBLISHER/$SERIES/$VOLUME/$SERIES v$VOLUME #$ISSUE [$COVERDATE]';
 
   let adaptor: LibraryAdaptor;
   let store: Store<AppState>;
@@ -312,12 +315,16 @@ describe('LibraryAdaptor', () => {
 
   describe('consolidating the library', () => {
     beforeEach(() => {
-      adaptor.consolidate(true);
+      adaptor.consolidate(true, DIRECTORY, RENAMING_RULE);
     });
 
     it('fires an action', () => {
       expect(store.dispatch).toHaveBeenCalledWith(
-        new LibraryConsolidate({ deletePhysicalFiles: true })
+        new LibraryMoveComics({
+          deletePhysicalFiles: true,
+          directory: DIRECTORY,
+          renamingRule: RENAMING_RULE
+        })
       );
     });
 
@@ -343,9 +350,7 @@ describe('LibraryAdaptor', () => {
             readingLists: []
           })
         );
-        store.dispatch(
-          new LibraryConsolidated({ deletedComics: DELETED_COMICS })
-        );
+        store.dispatch(new LibraryComicsMoved());
       });
 
       it('provides updates on consolidating', () => {
@@ -353,19 +358,11 @@ describe('LibraryAdaptor', () => {
           expect(response).toBeFalsy()
         );
       });
-
-      it('provides updates on comics', () => {
-        DELETED_COMICS.forEach(comic => {
-          adaptor.comic$.subscribe(comics =>
-            expect(comics).not.toContain(comic)
-          );
-        });
-      });
     });
 
     describe('failure', () => {
       beforeEach(() => {
-        store.dispatch(new LibraryConsolidateFailed());
+        store.dispatch(new LibraryMoveComicsFailed());
       });
 
       it('provides updates on consolidating', () => {
--- a/comixed-frontend/src/app/library/adaptors/library.adaptor.ts
+++ b/comixed-frontend/src/app/library/adaptors/library.adaptor.ts
@@ -31,10 +31,10 @@ import { extractField } from 'app/library/library.functions';
 import { LastReadDate } from 'app/library/models/last-read-date';
 import {
   LibraryClearImageCache,
-  LibraryConsolidate,
   LibraryConvertComics,
   LibraryDeleteMultipleComics,
   LibraryGetUpdates,
+  LibraryMoveComics,
   LibraryReset,
   LibraryStartRescan
 } from 'app/library/actions/library.actions';
@@ -300,12 +300,20 @@ export class LibraryAdaptor {
     return this._converting$.asObservable();
   }
 
-  consolidate(deletePhysicalFiles: boolean): void {
+  consolidate(
+    deletePhysicalFiles: boolean,
+    targetDirectory: string,
+    renamingRule: string
+  ): void {
     this.logger.debug(
       `firing action to consolidate library: deletePhysicalFiles=${deletePhysicalFiles}`
     );
     this.store.dispatch(
-      new LibraryConsolidate({ deletePhysicalFiles: deletePhysicalFiles })
+      new LibraryMoveComics({
+        deletePhysicalFiles: deletePhysicalFiles,
+        directory: targetDirectory,
+        renamingRule: renamingRule
+      })
     );
   }
 
--- a/comixed-frontend/src/app/library/components/consolidate-library/consolidate-library.component.ts
+++ b/comixed-frontend/src/app/library/components/consolidate-library/consolidate-library.component.ts
@@ -17,11 +17,15 @@
  */
 
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { FormBuilder, FormGroup } from '@angular/forms';
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { LoggerService } from '@angular-ru/logger';
 import { LibraryAdaptor } from 'app/library';
 import { Subscription } from 'rxjs';
-import { CONSOLIDATE_DELETE_PHYSICAL_FILES } from 'app/user/models/preferences.constants';
+import {
+  MOVE_COMICS_DELETE_PHYSICAL_FILE,
+  MOVE_COMICS_RENAMING_RULE,
+  MOVE_COMICS_TARGET_DIRECTORY
+} from 'app/user/models/preferences.constants';
 import { AuthenticationAdaptor } from 'app/user';
 import { ConfirmationService } from 'primeng/api';
 import { TranslateService } from '@ngx-translate/core';
@@ -47,7 +51,9 @@ export class ConsolidateLibraryComponent implements OnInit, OnDestroy {
     private translateService: TranslateService
   ) {
     this.consolidationForm = this.formBuilder.group({
-      deletePhysicalFiles: ['']
+      deletePhysicalFiles: [''],
+      targetDirectory: ['', Validators.required],
+      renamingRule: ['']
     });
     this.consolidatingSubscription = this.libraryAdaptor.consolidating$.subscribe(
       consolidating => (this.consolidating = consolidating)
@@ -55,7 +61,7 @@ export class ConsolidateLibraryComponent implements OnInit, OnDestroy {
     this.userSubscription = this.authenticationAdaptor.user$.subscribe(() => {
       this.consolidationForm.controls['deletePhysicalFiles'].setValue(
         this.authenticationAdaptor.getPreference(
-          CONSOLIDATE_DELETE_PHYSICAL_FILES
+          MOVE_COMICS_DELETE_PHYSICAL_FILE
         ) === '1'
       );
     });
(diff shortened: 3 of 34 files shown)


comixed-frontend/src/app/library/adaptors/library.adaptor.spec.ts
```diff
@@ -312,12 +315,16 @@ describe('LibraryAdaptor', () => {
 
   describe('consolidating the library', () => {
```
BRUCELLA2, line 316: Is that the describe should not be updated with 'moving the library' ?

comixed-frontend/src/app/library/adaptors/library.adaptor.ts
```diff
@@ -300,12 +300,20 @@ export class LibraryAdaptor {
     return this._converting$.asObservable();
   }
 
-  consolidate(deletePhysicalFiles: boolean): void {
+  consolidate(
```
BRUCELLA2, line 303: Is that the term 'consolidate' should not become 'move'? (if yes, the logger message need to be updated too).

comixed-frontend/src/app/library/components/consolidate-library/consolidate-library.component.ts
```diff
@@ -47,15 +51,17 @@ export class ConsolidateLibraryComponent implements OnInit, OnDestroy {
...
     });
     this.consolidatingSubscription = this.libraryAdaptor.consolidating$.subscribe(
       consolidating => (this.consolidating = consolidating)
     );
     this.userSubscription = this.authenticationAdaptor.user$.subscribe(() => {
       this.consolidationForm.controls['deletePhysicalFiles'].setValue(
         this.authenticationAdaptor.getPreference(
-          CONSOLIDATE_DELETE_PHYSICAL_FILES
+          MOVE_COMICS_DELETE_PHYSICAL_FILE
         ) === '1'
```
BRUCELLA2, line 65: Reading the code I find that this preference can be '1' or '0' but it's not easy to understand the meaning of this value. I think we use strings instead of booleans to make it work with all preferences.
I think we use strings instead of booleans to make it work with all preferences like 'TRUE' or 'FALSE' with the value 1 or 0 or something like this ?

comixed-rest-api/src/main/java/org/comixed/net/MoveComicsRequest.java
```diff
@@ -0,0 +1,51 @@
...
+    return deletePhysicalFiles;
+  }
+
+  public String getTargetDirectory() {
+    return targetDirectory;
+  }
+
+  public String getRenamingRule() {
+    return renamingRule;
+  }
```
BRUCELLA2, line 50: Wouldn't it be useful to use lombok annotations in this case?
With the @AllArgsConstructor and @Getter annotations if we want the constructor with all arguments and only the getters. And @AllArgsConstructor and @Data if we want to generate the constructor with all arguments and the getters / setters / tostring / equals / hashcode 
if annotations are not used, wouldn't it be a good idea to make the parameters 'final'?

comixed-rest-api/src/main/java/org/comixed/controller/library/LibraryController.java
```diff
@@ -166,4 +166,20 @@ public ClearImageCacheResponse clearImageCache() {
 
     return new ClearImageCacheResponse(true);
   }
+
+  @PostMapping(
+      value = "/library/move",
+      produces = MediaType.APPLICATION_JSON_VALUE,
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public MoveComicsResponse moveComics(@RequestBody() MoveComicsRequest request) {
```
BRUCELLA2, line 174: Wouldn't it be nice to make the parameters 'final'?

comixed-rest-api/src/main/java/org/comixed/controller/library/LibraryController.java
```diff
@@ -166,4 +166,20 @@ public ClearImageCacheResponse clearImageCache() {
...
+    String targetDirectory = request.getTargetDirectory();
+    String renamingRule = request.getRenamingRule();
+    Boolean deletePhysicalFiles = request.getDeletePhysicalFiles();
```
BRUCELLA2, line 177: Wouldn't it be a good idea to make the variables 'final' since they are not intended to be re-assigned?

comixed-rest-api/src/main/java/org/comixed/net/MoveComicsResponse.java
```diff
@@ -0,0 +1,34 @@
...
+  @JsonProperty("success")
+  private boolean success;
+
+  public MoveComicsResponse(boolean success) {
+    this.success = success;
+  }
+
+  public boolean isSuccess() {
+    return this.success;
+  }
```
BRUCELLA2, line 33: As before, couldn't we use the lombok annotations?

comixed-services/src/main/java/org/comixed/service/library/LibraryService.java
```diff
@@ -164,4 +166,22 @@ public void clearImageCache() throws LibraryException {
...
   }
+
+  /**
+   * Moves all comics in the library, renaming them using the specified naming rule.
+   *
+   * @param deletePhysicalFiles
+   * @param directory the root directory
+   * @param renamingRule the name rule
+   */
+  public void moveComics(Boolean deletePhysicalFiles, String directory, String renamingRule) {
```
BRUCELLA2, line 177: Wouldn't it be a good idea to make the parameters 'final' since they are not intended to be reallocated?

comixed-tasks/src/main/java/org/comixed/task/encoders/MoveComicTaskEncoder.java
```diff
@@ -0,0 +1,109 @@
...
+public class MoveComicTaskEncoder extends AbstractTaskEncoder<MoveComicWorkerTask> {
+  public static final String DIRECTORY = "destination-directory";
+  public static final String RENAMING_RULE = "renaming-rule";
+
+  @Autowired private ObjectFactory<MoveComicWorkerTask> moveComicWorkerTaskObjectFactory;
+  @Autowired private TaskRepository taskRepository;
+
+  private Comic comic;
+  private String directory;
+  private String renamingRule;
```
BRUCELLA2, line 50: Couldn't we use the lombok annotations?

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Sets the renaming rule to be used.
+   *
+   * @param renamingRule the renaming rule
+   */
+  public void setRenamingRule(String renamingRule) {
+    this.renamingRule = renamingRule;
+  }
```
BRUCELLA2, line 97: Couldn't we use the lombok annotations?

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Autowired private TaskRepository taskRepository;
+  @Autowired private ComicRepository comicRepository;
+
+  private String directory;
+  private String renamingRule;
+
+  @Override
+  protected String createDescription() {
+    return "Moving comics";
```
BRUCELLA2, line 56: Couldn't we have a constant string?

comixed-tasks/src/test/resources/application.properties
```diff
@@ -20,7 +20,7 @@ spring.jackson.deserialization.fail-on-unknown-properties=false
 
 # Logging
 logging.level.root=OFF
-logging.level.org.comixed=INFO
+logging.level.org.comixed=DEBUG
```
BRUCELLA2, line 23: Is pushing this change is voluntary? (in production debug level can impact performance)

BRUCELLA2 (commented):
Some suggestions for changes and some questions... But on the whole I don't see anything that prevents the merging.

comixed-frontend/src/app/library/adaptors/library.adaptor.spec.ts
mcpierce: I went back and forth on it. Ultimately it's still consolidation since you don't have to move the comics to a new directory to use this feature: it's going to be the way you convert comics, delete comics marked for removal, etc. as well as moving them.

comixed-frontend/src/app/library/adaptors/library.adaptor.ts
mcpierce: See previous response.

comixed-frontend/src/app/library/components/consolidate-library/consolidate-library.component.ts
mcpierce: I created #387 to track making this change.

comixed-rest-api/src/main/java/org/comixed/net/MoveComicsRequest.java
mcpierce: The code uses the explicit constructor so that these objects can be created as part of the JUnit tests, where different request objects are created to test different scenarios. Are those lombok annotations usable from JUnit as well?

comixed-rest-api/src/main/java/org/comixed/controller/library/LibraryController.java
mcpierce: It would be, yes. I think we've got a mixed bag of final and non-final parameters in the codebase. We should probably have an issue created to do this across the board. I want to add REST Assured unit tests to the REST APIs and I think we could standardize things as a part of that effort. There's a story (#365) for this effort.

comixed-services/src/main/java/org/comixed/service/library/LibraryService.java
mcpierce: See previous response.

comixed-tasks/src/main/java/org/comixed/task/encoders/MoveComicTaskEncoder.java
mcpierce: Yeah, let's open a new story to do that across the board. I created #386 to track this.

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
mcpierce: See previous answer.

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
mcpierce: I think constants are better when the same value is used in more than one spot. If it's a single point, like here, they don't really give any benefit.

comixed-tasks/src/test/resources/application.properties
mcpierce: Yes, it was. This is only used as a part of the JUnit test run: the file's in ```src/test/resources```.

comixed-frontend/src/app/library/adaptors/library.adaptor.spec.ts
BRUCELLA2: ok i understand.

comixed-frontend/src/app/library/adaptors/library.adaptor.ts
BRUCELLA2: ok

comixed-frontend/src/app/library/components/consolidate-library/consolidate-library.component.ts
BRUCELLA2: ok :)

comixed-rest-api/src/main/java/org/comixed/net/MoveComicsRequest.java
BRUCELLA2: Yes lombok annotations are usable from Junit.

comixed-rest-api/src/main/java/org/comixed/controller/library/LibraryController.java
BRUCELLA2: Yes it's good idea.

comixed-services/src/main/java/org/comixed/service/library/LibraryService.java
BRUCELLA2: ok

comixed-tasks/src/main/java/org/comixed/task/encoders/MoveComicTaskEncoder.java
BRUCELLA2: ok nice

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
BRUCELLA2: ok

comixed-tasks/src/main/java/org/comixed/task/model/MoveComicsWorkerTask.java
BRUCELLA2: yes don't give more benefit ok

comixed-tasks/src/test/resources/application.properties
BRUCELLA2: I didn't see it was for the tests :)

BRUCELLA2 (approved):
ok for me. I approve this PR

mcpierce:
@BRUCELLA2 When you approve the PR, you can do the "Rebase and Merge" button on it so that the code gets merged into develop and the PR closed.

BRUCELLA2:
I think I need some write access to Rebase and Merge.

mcpierce:
You have to start a review first. Take a look at [this guide](https://docs.github.com/en/enterprise/2.13/user/articles/reviewing-proposed-changes-in-a-pull-request) for it. Basically, you start the review when adding a comment, then when you're done you can either mark the review as requiring changes or you can approve it. If the latter, then you can go ahead and do the merge.

BRUCELLA2:
I've already done the review  and approve it :) 


And the merging is blocked for me :

mcpierce:
@BRUCELLA2 Try now -- I changed your access so you should be able to do it.

BRUCELLA2 (approved):

BRUCELLA2:
It's good

PR #388, opened 2020-07-01
Users can set a maximum number of comics to fetch during import

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Added a request parameter to set the maximum number of comic files to return. Added a dropdown to the frontend to let the user set a persistent maximum value, and to submit it during import.

### DIFF
--- a/comixed-rest-api/src/main/java/org/comixed/net/GetAllComicsUnderRequest.java
+++ b/comixed-rest-api/src/main/java/org/comixed/net/GetAllComicsUnderRequest.java
@@ -20,21 +20,41 @@ package org.comixed.net;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+/**
+ * <code>GetAllComicsUnderRequest</code> represents the request body during comic file imports.
+ *
+ * @author Darryl L. Pierce
+ */
 public class GetAllComicsUnderRequest {
   @JsonProperty("directory")
   private String directory;
 
+  @JsonProperty("maximum")
+  private Integer maximum;
+
   public GetAllComicsUnderRequest() {}
 
-  public GetAllComicsUnderRequest(final String directory) {
+  public GetAllComicsUnderRequest(final String directory, final Integer maximum) {
     this.directory = directory;
+    this.maximum = maximum;
   }
 
+  /**
+   * The root directory to be searched.
+   *
+   * @return the directory
+   */
   public String getDirectory() {
     return directory;
   }
 
-  public void setDirectory(final String directory) {
-    this.directory = directory;
+  /**
+   * The maximum number of comics to find. If the value is zero then then all comics found are
+   * returned.
+   *
+   * @return the maximum
+   */
+  public Integer getMaximum() {
+    return maximum;
   }
 }
--- a/comixed-services/src/main/java/org/comixed/service/file/FileService.java
+++ b/comixed-services/src/main/java/org/comixed/service/file/FileService.java
@@ -77,7 +77,8 @@ public class FileService {
     return result;
   }
 
-  public List<FileDetails> getAllComicsUnder(final String rootDirectory) throws IOException {
+  public List<FileDetails> getAllComicsUnder(final String rootDirectory, final Integer maximum)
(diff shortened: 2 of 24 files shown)


mcpierce:
I'm going to fix the three code smells before we merge this.

mcpierce:
Okay, I've fixed the issues SonarCloud found.

comixed-rest-api/src/main/java/org/comixed/net/GetAllComicsUnderRequest.java
```diff
@@ -20,21 +20,41 @@
...
+   */
   public String getDirectory() {
     return directory;
   }
 
-  public void setDirectory(final String directory) {
-    this.directory = directory;
+  /**
+   * The maximum number of comics to find. If the value is a non-positive valume then all comics
+   * found are returned.
```
BRUCELLA2: 0 is a positive value in this context ? In French mathematical zero is both positive and negative : in mathematical French "x est positif" (literally "x is positive") allows the case x=0, while "x est positif strictement" (literally "x is strictly positive") does not. There are many other examples in mathematics where "positive" is used to include  0. 
To be clearer, if 0 is not a positive value in this context, it would be better to say if the value is < 1 all comics found will be returned.

comixed-services/src/main/java/org/comixed/service/file/FileService.java
```diff
@@ -77,15 +77,16 @@
     return result;
   }
 
-  public List<FileDetails> getAllComicsUnder(final String rootDirectory) throws IOException {
+  public List<FileDetails> getAllComicsUnder(final String rootDirectory, Integer maximum)
```
BRUCELLA2: it's better if maximum is final

comixed-services/src/main/java/org/comixed/service/file/FileService.java
```diff
@@ -96,33 +97,40 @@
     return result;
   }
 
-  private void loadFilesUnder(final List<FileDetails> files, final File directory)
+  private void loadFilesUnder(final List<FileDetails> files, final File directory, Integer maximum)
```
BRUCELLA2: it's better if maximum is final

comixed-services/src/main/java/org/comixed/service/file/FileService.java
```diff
@@ -96,33 +97,40 @@
...
-              files.add(new FileDetails(filePath, fileSize));
-            }
+            files.add(new FileDetails(filePath, fileSize));
           }
         }
       }
     }
   }
 
+  private boolean canBeImported(File file) throws IOException {
```
BRUCELLA2: It's better if file is final

BRUCELLA2 (changes requested):
Some modifications concerning small details may be necessary.

comixed-rest-api/src/main/java/org/comixed/net/GetAllComicsUnderRequest.java
mcpierce: No. If the value is greater than zero then it's setting a limit on the number of results to return. If the value is zero then there is **no** limit.  The Javadoc may be less than clear in this case.

mcpierce: Changed the Javadoc.

comixed-services/src/main/java/org/comixed/service/file/FileService.java
mcpierce: Fixed.

comixed-services/src/main/java/org/comixed/service/file/FileService.java
mcpierce: Fixed.

comixed-services/src/main/java/org/comixed/service/file/FileService.java
mcpierce: Fixed.

BRUCELLA2 (approved):

PR #394, opened 2020-07-05
Deleted pages are removed from comics when they're converted.

mcpierce:
## Status
**READ**

## Migrations
NO

## Description
Enhanced the REST APIs to allow deleting pages that are either blocked or marked for deletion when the comics are converted or re-archived..

### DIFF
--- a/comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
+++ b/comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
@@ -20,36 +20,44 @@ package org.comixed.net;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.List;
+import lombok.Getter;
+import lombok.Setter;
 import org.comixed.adaptors.ArchiveType;
 
+/**
+ * <code>ConvertComicsRequest</code> provides the HTTP request contents when converting comics.
+ *
+ * @author Darryl L. Pierce
+ */
 public class ConvertComicsRequest {
+  @Getter
+  @Setter
   @JsonProperty("ids")
   private List<Long> comicIdList;
 
+  @Getter
+  @Setter
   @JsonProperty("archiveType")
   private ArchiveType archiveType;
 
+  @Getter
+  @Setter
   @JsonProperty("renamePages")
   private boolean renamePages;
 
-  public ConvertComicsRequest() {}
+  @Getter
+  @Setter
+  @JsonProperty("deletePages")
+  private boolean deletePages;
 
   public ConvertComicsRequest(
-      List<Long> comicIdList, ArchiveType archiveType, boolean renamePages) {
+      final List<Long> comicIdList,
+      final ArchiveType archiveType,
+      final boolean renamePages,
+      final Boolean deletePages) {
     this.comicIdList = comicIdList;
     this.archiveType = archiveType;
     this.renamePages = renamePages;
-  }
-
-  public List<Long> getComicIdList() {
-    return comicIdList;
-  }
-
-  public ArchiveType getArchiveType() {
-    return archiveType;
-  }
-
-  public boolean getRenamePages() {
-    return renamePages;
+    this.deletePages = deletePages;
   }
 }
--- a/.github_changelog_generator
+++ b/.github_changelog_generator
@@ -1,0 +1,1 @@
+enhancement_labels=feature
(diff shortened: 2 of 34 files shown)


comixed-frontend/src/app/library/library.constants.ts
```diff
@@ -16,9 +16,9 @@
  * along with this program. If not, see <http://www.gnu.org/licenses>
  */
 
-import { API_ROOT_URL } from 'app/app.functions';
-import { COMIXED_API_ROOT } from 'app/app.constants';
-import { ReadingList } from 'app/comics/models/reading-list';
+import {API_ROOT_URL} from 'app/app.functions';
+import {COMIXED_API_ROOT} from 'app/app.constants';
+import {ReadingList} from 'app/comics/models/reading-list';
```
BRUCELLA2: I thought prettier was adding a space after the opening brace and the ending brace. Could you check this ?

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
```diff
@@ -20,36 +20,44 @@
...
 import org.comixed.adaptors.ArchiveType;
 
+/**
+ * <code>ConvertComicsRequest</code> provides the HTTP request contents when converting comics.
+ *
+ * @author Darryl L. Pierce
+ */
 public class ConvertComicsRequest {
+  @Getter
+  @Setter
```
BRUCELLA2, line 34: If you want the getter and the setter for all the fields, this annotations can be put on the class.

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
```diff
@@ -20,36 +20,44 @@
...
-    return comicIdList;
-  }
-
-  public ArchiveType getArchiveType() {
-    return archiveType;
-  }
-
-  public boolean getRenamePages() {
-    return renamePages;
+    this.deletePages = deletePages;
```
BRUCELLA2, line 61: For this constructor you can use the annotation @AllArgsConstructor on the class.

BRUCELLA2 (changes requested):
I think that some tests are missing in particular for the ConvertComicWorkerTask class and the Comic class.
The code smell in  ConvertComicTaskEncoder  (Rename "targetArchiveType" which hides the field declared at line 49.) need to be fix.

comixed-frontend/src/app/library/library.constants.ts
mcpierce: Correct, I'll go back and fix this and repush.

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
mcpierce: I would prefer to be specific with annotations like this, to avoid accidentally exposing a state variable that shouldn't be exposed. Annotating each field specifically helps make it explicit that it should be exposed with either a getter and/or a setter.

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
mcpierce: I don't think that will work since this represents an HTTP request payload: when it gets created in a real-life transaction, I believe Spring only uses the no-arg constructor. So it wouldn't be able to set the fields if they're marked as final, which is how I read Lombok's docs for the AllArgsConstructor annotation.

Is that amistaken understanding on my part?

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
BRUCELLA2: It's good. I prefer to write more code to be specific and have my intentions clear, rather than writing less and leaving gaps that could lead to errors later :)

comixed-rest-api/src/main/java/org/comixed/net/ConvertComicsRequest.java
BRUCELLA2: I don't know if this will work... if i put the annotation @AllArgsConstructor and I delombok to see what lombok do i see this : 
  public ConvertComicsRequest(List<Long> comicIdList, ArchiveType archiveType, boolean renamePages)
  {
    this.comicIdList = comicIdList;
    this.archiveType = archiveType;
    this.renamePages = renamePages;
  }

Your own constructor seems to be better with the final keyword... I think we can continue to do without this annotation...

PR #397, opened 2020-07-06
Replace Worker with a thread pool manager

mcpierce:
## Status
**READ**

## Migrations
NO

## Description
Replaced Worker with TaskManager, which uses a ThreadPoolTaskExecutor to run the tasks in parallel. Removed the dependency on ConcurrentUnit as no longer needed.


BRUCELLA2 (approved):
I don't know much about thread pool managers but from what I've seen the pool request is approved.

Issue #333, opened 2020-06-10
Provide French language translations


mcpierce:
Provide a translation of the application in French.

PR #406, opened 2020-07-11
French translation [#333]

BRUCELLA2:
## Status
**READY**

## Migrations
NO

## Description
Added french translation

### DIFF
--- a/comixed-frontend/src/app/components/navigation-bar/navigation-bar.component.scss
+++ b/comixed-frontend/src/app/components/navigation-bar/navigation-bar.component.scss
@@ -9,3 +9,7 @@
   color: $cx-toolbar-text-color;
   margin-right: 2px;
 }
+
+::ng-deep .ui-menu {
+  width: max-content;
+}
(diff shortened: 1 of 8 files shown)


mcpierce (changes requested):
Can you squash this commit into the other so it's a single commit?

mcpierce:
This all looks good. Once the second commit is squashed into the first I'll merge it.

BRUCELLA2:
Squash done

mcpierce (commented):
I'm having trouble getting the translation to load. I believe there's an issue in one of the interpolated values, but haven't figured out which one yet. I see this error when I switch to French:

```
core.js:4002 ERROR n {message: "Expected "=", "}", or identifier but end of input found.", expected: Array(3), found: null, location: {…}, name: "SyntaxError", …}
```

but haven't determined which entry it is.

Are you able to use the French translation locally?

BRUCELLA2:
I found the mistake. I also adjusted the size of the main menu to the size of the text (it was out of range with the French translation).

mcpierce (approved):
Looks great and works on my system.

PR #411, opened 2020-07-17
Replace the old scraping code with new WebFlux-based scraping actions

mcpierce:
## Status
**READY**

## Migrations
YES

## Description
Designed a new scraping system based on a ScrapingAdaptor and individual ScrapingAction objects. Replaced the old caching tables with a new concise pair of tables.


comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Returns a list of volumes for the given series name.
+   *
+   * @param apiKey the API key
+   * @param seriesName the series name
+   * @param skipCache the skip cache flag
+   * @return the list of volumes
+   * @throws ScrapingException if an error occurs
+   */
+  List<ScrapingVolume> getVolumes(String apiKey, String seriesName, boolean skipCache)
```
BRUCELLA2, line 43: Parameters can be declared "final" (they are final in the implementation).

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
```diff
@@ -0,0 +1,73 @@
...
+   * Returns a single issue.
+   *
+   * @param apiKey the API key
+   * @param volume the volume
+   * @param issueNumber the issue number within the volume
+   * @param skipCache the skip cache flag
+   * @return the issue or null
+   * @throws ScrapingException if an error occurs
+   */
+  ScrapingIssue getIssue(String apiKey, Integer volume, String issueNumber, boolean skipCache)
```
BRUCELLA2, line 56: Parameters can be declared "final" (they are final in the implementation).

comixed-scrapers/src/main/java/org/comixed/scrapers/comicvine/actions/ComicVineGetIssueDetailsAction.java
```diff
@@ -0,0 +1,83 @@
...
+    this.addField("location_credits");
+    this.addField("story_arc_credits");
+    this.addField("person_credits");
+
+    if (StringUtils.isEmpty(this.getApiKey())) throw new ScrapingException("Missing API key");
+    if (StringUtils.isEmpty(this.issueId)) throw new ScrapingException("Missing issue id");
+
+    log.debug(
+        "Querying ComicVine for issue: id={} API key={}", this.issueId, this.getMaskedApiKey());
+    final String url = this.createUrl(this.baseUrl, this.getEndpoint());
```
BRUCELLA2, line 68: if baseUrl is null couldn't that be a problem?

BRUCELLA2 (changes requested):
3 minor details to check.

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
mcpierce: Isn’t the final keyword only useful on an implementation: ie an implementing class can drop the final keyword?

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
mcpierce: See previous reply.

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
BRUCELLA2: I learn something :) I'm still a rookie ;)
It's true, final can be drop in the implementation and adding the final word in an interface can therefore be counter-intuitive as the interface contract that indicates that the parameter will not be reassigned could be reassigned in the implementation.
Don't you think then that we should remove the "final" in the scrapeComic method in the interface?

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
BRUCELLA2: ok see previous reply.

comixed-scrapers/src/main/java/org/comixed/scrapers/adaptors/ScrapingAdaptor.java
mcpierce: That's what I was getting at before: "final" only applies to an actual implementation of the method and not to the interface. We could put "final" here and someone could implement the interface and leave it off and it wouldn't be a compile time error: final serves no purpose in an interface.

comixed-scrapers/src/main/java/org/comixed/scrapers/comicvine/actions/ComicVineGetIssueDetailsAction.java
mcpierce: Yep, thanks for catching that. I've updated the branch to include at test for that condition.

BRUCELLA2 (approved):
it's ok for me

Issue #415, opened 2020-07-17
Update scraping dialog title and message


mcpierce:
The label should read: **Update Comic Details**
The message should read: **Are you sure you want to replace the comic's details?**

PR #416, opened 2020-07-17
Changed the scraping confirmation dialog title and message [#415]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Changed the english language text for the scraping confirmation dialog box.

### DIFF
--- a/comixed-frontend/src/assets/i18n/en/app.json
+++ b/comixed-frontend/src/assets/i18n/en/app.json
@@ -164,9 +164,6 @@
       "preferences": "Preferences"
     }
   },
-  "users-page": {
-    "title": "ComiXed: Manager User Accounts"
-  },
   "comic-list": {
     "label": {
       "series": "Series",
--- a/comixed-frontend/src/assets/i18n/en/comics.json
+++ b/comixed-frontend/src/assets/i18n/en/comics.json
@@ -231,8 +231,8 @@
       "message": "Are you sure you want to save the changes made to this comic?"
     },
     "issue-selected": {
-      "header": "Scrape This Issue?",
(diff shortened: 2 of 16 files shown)


BRUCELLA2 (changes requested):
The Spanish and French translations also need to be updated, don't they? 
(It's not related with this PR but I see some tags are missing in french and spanish comics.json files)

mcpierce:
> The Spanish and French translations also need to be updated, don't they?
> (It's not related with this PR but I see some tags are missing in french and spanish comics.json files)

No, looking at them they didn't appear to be translating the word "scrape" that I could tell.

Which tags are missing? I can fix that with this pr to make things complete.

BRUCELLA2:
In the comics.json in french and spanish, the tag comics-pages is not the same as in english (lines 173 - 182 in french/spanish and lines 173 - 203 in english).
Other things, the tages main-menu -> item -> account is 2 times in common.json in all language.
And the tag users-page -> title and users-page -> text have to be put together i think.

For the translation to update in french and in spanish we don't use the scape word but we use the reference to ComicVine and if we want something generic we can update this.

mcpierce:
> In the comics.json in french and spanish, the tag comics-pages is not the same as in english (lines 173 - 182 in french/spanish and lines 173 - 203 in english).
> Other things, the tages main-menu -> item -> account is 2 times in common.json in all language.
> And the tag users-page -> title and users-page -> text have to be put together i think.
> 
> For the translation to update in french and in spanish we don't use the scape word but we use the reference to ComicVine and if we want something generic we can update this.

Very good. I've gone ahead and addressed those issues in this PR.

BRUCELLA2 (approved):

PR #436, opened 2020-07-24
Moved models, adaptors and repositories to their own submodules.

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Moved all of the components from comixed-library to their own modules. Refactored the model objects to use lombok annotations for fields. Moved a lot (but not all) business logic out of the models into adaptors. Refactored the remaining code to use the updated APIs.


comixed-model/src/main/java/org/comixedproject/model/user/Bookmark.java
```diff
@@ -20,6 +20,8 @@
 
 import com.fasterxml.jackson.annotation.JsonView;
 import javax.persistence.*;
+import lombok.Getter;
+import lombok.Setter;
 import org.comixedproject.model.comic.Comic;
 import org.comixedproject.views.View.UserList;
```
BRUCELLA2, line 27: Comments on the class are missing (like @author ...)

comixed-model/src/main/java/org/comixedproject/model/user/Preference.java
```diff
@@ -40,14 +40,18 @@
   @ManyToOne
   @JoinColumn(name = "user_id")
   @JsonIgnore
+  @Getter
   private ComiXedUser user;
 
   @Column(name = "name", nullable = false)
   @JsonView(UserList.class)
+  @Getter
```
BRUCELLA2, line 48: In last version of this class there is a setName, is it voluntary not to put it here?

BRUCELLA2 (commented):
Code coverage is less than 80%. Many lines not covered are in model with lombok generated code. What do you think about adding a configuration that would exclude the code generated by lombok from the coverage to be done? This can be done by adding a lombok.config file with the parameters : 
config.stopBubbling = true
lombok.addLombokGeneratedAnnotation = true

- config.stopBubbling = true tells Lombok that this is your root directory and that it should not search parent directories for more configuration files (you can have more than one lombok config files in different directories/packages).

- lombok.addLombokGeneratedAnnotation = true will add @lombok.Generated annotation to all Lombok generated methods.

comixed-model/src/main/java/org/comixedproject/model/user/Bookmark.java
mcpierce: > config.stopBubbling = true
> lombok.addLombokGeneratedAnnotation = true

Done. I assume this file goes in the root of the project?

mcpierce: Fixed, added a Javadoc for the model and added Joao Franca as the original author.

comixed-model/src/main/java/org/comixedproject/model/user/Preference.java
mcpierce: Yes. I think it's more consistent to, in cases where it's a collection, to have the caller get the collection and use its add method. That way we make the models simpler. I also did the same for things like the characters/teams/locations/stories on Comic, for example.

BRUCELLA2: Yes it's better.

comixed-model/src/main/java/org/comixedproject/model/user/Bookmark.java
BRUCELLA2: Yes, the file goes in the root of the project

BRUCELLA2 (approved):

PR #449, opened 2020-07-28
Remove page type from the system

mcpierce:
## Status
**READY**

## Migrations
YES

## Description
Dropped the page_types table and all references to it. Dropped the type_id column from pages. Removed PageTypeRepository and all REST APIs that referenced it. Dropped the page type actions from the frontend and UX.


mcpierce:
Closing this PR since we're going to wait and see if page types will be useful in future.

PR #454, opened 2020-08-01
Changed the REST API layer to initiate Tasks

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Reworked the REST API layer to create tasks as needed. Added the ConvertComicsWorkerTask.


comixed-repositories/src/main/java/org/comixedproject/repositories/comic/ComicRepository.java
```diff
@@ -109,4 +109,7 @@
 
   @Query("SELECT c FROM Comic c")
   List<Comic> findComicsToMove(PageRequest page);
+
+  @Query("SELECT c FROM Comic c WHERE c.id > :threshold ORDER BY c.id")
+  List<Comic> getComicsById(@Param("threshold") Long threshold, Pageable page);
```
BRUCELLA2: I'm not sure the name of the method is representative of what it does. by reading the name getComicsById, I think we're going to provide as a parameter a list of id's and receive a list of comics matching that list. Whereas it seems that we get a list of comics whose ids are above a threshold... What do you think about naming the method: getComicsByIdThreshold ?

BRUCELLA2: With javadoc comment is missing too

comixed-services/src/test/java/org/comixedproject/service/comic/PageServiceTest.java
```diff
@@ -531,4 +532,16 @@ public void testSetDeletedStateUnmark() {
     Mockito.verify(page, Mockito.times(1)).setDeleted(false);
     Mockito.verify(pageRepository, Mockito.times(1)).save(page);
   }
+
+  @Test
+  public void testDelete() {
```
BRUCELLA2: isn't testSave() no ?

comixed-tasks/src/main/java/org/comixedproject/task/model/ConvertComicsWorkerTask.java
```diff
@@ -25,53 +25,65 @@
...
-      log.debug("Queueing task to save comic: id={}", comic.getId());
-      ConvertComicTaskEncoder encoder = this.saveComicTaskEncoderObjectFactory.getObject();
+    for (Long id : this.idList) {
+      log.debug("Queueing task to save comic: id={}", id);
+      Comic comic = null;
+      try {
+        comic = this.comicService.getComic(id);
+      } catch (ComicException error) {
+        throw new WorkerTaskException("failed to load comic", error);
+      }
```
BRUCELLA2, line 78: an enhanced loop would be more readable

comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicsWorkerTask.java
```diff
@@ -51,15 +60,20 @@ protected String createDescription() {
   public void startTask() throws WorkerTaskException {
     for (int index = 0; index < this.comicIds.size(); index++) {
       final Long id = this.comicIds.get(index);
-      final Comic comic = this.comicRepository.getById(id);
+      Comic comic;
+      try {
+        comic = this.comicService.getComic(id);
+      } catch (ComicException error) {
+        throw new WorkerTaskException("failed to load comic", error);
+      }
```
BRUCELLA2, line 68: an enhanced loop would be more readable

comixed-tasks/src/main/java/org/comixedproject/task/model/RescanComicsWorkerTask.java
```diff
@@ -0,0 +1,81 @@
...
+      if (!comics.isEmpty()) {
+        for (int index = 0; index < comics.size(); index++) {
+          final Comic comic = comics.get(index);
+          final RescanComicWorkerTaskEncoder encoder =
+              this.rescanComicWorkerTaskEncoderObjectFactory.getObject();
+          encoder.setComic(comic);
+          this.taskService.save(encoder.encode());
+          if (comic.getId() > lastId) lastId = comic.getId();
+          count++;
+        }
```
BRUCELLA2: an enhanced loop would be more readable

comixed-tasks/src/main/java/org/comixedproject/task/model/RescanComicsWorkerTask.java
```diff
@@ -0,0 +1,81 @@
...
+  @Autowired
+  private ObjectFactory<RescanComicWorkerTaskEncoder> rescanComicWorkerTaskEncoderObjectFactory;
+
+  @Override
+  protected String createDescription() {
+    return "Rescanning comics";
+  }
+
+  @Override
+  public void startTask() throws WorkerTaskException {
```
BRUCELLA2, line 56: WorkerTaskException is never thrown in the method.

BRUCELLA2 (changes requested):

BRUCELLA2:
There is some conficting files too.

comixed-repositories/src/main/java/org/comixedproject/repositories/comic/ComicRepository.java
mcpierce: Renamed this method to *findComicsWithIdGreaterThan* and added Javadoc.

comixed-tasks/src/main/java/org/comixedproject/task/model/ConvertComicsWorkerTask.java
mcpierce: Not sure I follow: isn't this an enhanced for loop? Or is this an old comment and I didn't mark i as resolved? :smile:

comixed-tasks/src/main/java/org/comixedproject/task/model/RescanComicsWorkerTask.java
mcpierce: Fixed.

comixed-tasks/src/main/java/org/comixedproject/task/model/ConvertComicsWorkerTask.java
BRUCELLA2: yes it's an enhanced loop :D

comixed-tasks/src/main/java/org/comixedproject/task/model/RescanComicsWorkerTask.java
mcpierce: But the method signature that's being overridden does. To keep things consistent, I don't think we should change the signature at the implementation just because it doesn't actually throw the exception.

comixed-services/src/test/java/org/comixedproject/service/comic/PageServiceTest.java
mcpierce: Yes, this is fixed.

comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicsWorkerTask.java
mcpierce: Yes, but I believe we'd introduce the ConcurrentModificationException issue with JPA. So it's ultimately better to use an index to access the reading lists.

BRUCELLA2: Ok

comixed-tasks/src/main/java/org/comixedproject/task/model/RescanComicsWorkerTask.java
BRUCELLA2: ok it's true

BRUCELLA2 (approved):

PR #475, opened 2020-08-18
Add annotation to create audit log entries for REST APIs

mcpierce:
## Status
**READY**

## Migrations
YES

## Description
Introduces a new annotation for auditing REST API calls. It doesn't annotate all requests since they're going to each need to be refactored to return ApiResponse<?> objects, which the annotation will process.


mcpierce:
I'm going to do some additional work on this before pushing it.

PR #476, opened 2020-08-19
Added REST API logging

mcpierce:
## Status
**READY**

## Migrations
YES

## Description
Wraps annotated REST APIs and logs their runtime, request and response content, and any exceptions.

### DIFF
--- a/comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
+++ b/comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
@@ -1,0 +1,85 @@
+/*
+ * ComiXed - A digital comic book library management application.
+ * Copyright (C) 2020, The ComiXed Project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses>
+ */
+
+package org.comixedproject.model.auditlog;
+
+import java.util.Date;
+import javax.persistence.*;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * <code>RestAuditLogEntry</code> represents a single entry in the REST API audit log table.
+ *
+ * @author Darryl L. Pierce
+ */
+@Entity
+@Table(name = "rest_audit_log")
+public class RestAuditLogEntry {
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @Getter
+  private Long id;
+
+  @Column(name = "remote_ip", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private String remoteIp;
+
+  @Column(name = "url", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private String url;
+
+  @Column(name = "method", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private String method;
+
+  @Column(name = "content", nullable = true, updatable = false)
+  @Lob
+  @Getter
+  @Setter
+  private String content;
+
+  @Column(name = "email", nullable = true, updatable = false)
+  @Getter
+  @Setter
+  private String emai;
+
+  @Column(name = "start_time", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private Date startTime = new Date();
+
+  @Column(name = "end_time", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private Date endTime = new Date();
+
+  @Column(name = "successful", nullable = false, updatable = false)
+  @Getter
+  @Setter
+  private Boolean successful;
+
+  @Column(name = "exception", nullable = true, updatable = false)
+  @Lob
+  @Getter
+  @Setter
+  private String exception;
+}
--- a/comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
+++ b/comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
@@ -51,16 +51,22 @@ public class RestAuditLogEntry {
   @Setter
   private String method;
 
-  @Column(name = "content", nullable = true, updatable = false)
+  @Column(name = "request_content", nullable = true, updatable = false)
   @Lob
   @Getter
   @Setter
-  private String content;
+  private String requestContent;
+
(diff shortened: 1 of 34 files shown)


comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
```diff
@@ -0,0 +1,91 @@
...
+  @Column(name = "response_content", nullable = true, updatable = false)
+  @Lob
+  @Getter
+  @Setter
+  private String responseContent;
+
+  @Column(name = "email", nullable = true, updatable = false)
+  @Getter
+  @Setter
+  private String emai;
```
BRUCELLA2: I think "email" would be better ;)

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationProvider.java
```diff
@@ -48,10 +54,16 @@ public Authentication authenticate(Authentication authentication) throws Authent
 
     log.debug("Attempting to authenticate: email={}", email);
 
-    ComiXedUser user = userRepository.findByEmail(email);
+    ComiXedUser user = null;
```
BRUCELLA2, line 59: you don't have to initialize user to null unless you want to be explicit, no?

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedUserDetailsService.java
```diff
@@ -30,27 +32,37 @@
...
 public class ComiXedUserDetailsService implements UserDetailsService {
-  @Autowired private ComiXedUserRepository userRepository;
+  @Autowired private UserService userService;
 
   @Override
   public UserDetails loadUserByUsername(String email) throws UsernameNotFoundException {
     log.debug("Loading user: email={}", email);
 
-    ComiXedUser user = userRepository.findByEmail(email);
+    ComiXedUser user = null;
```
BRUCELLA2, line 48: Like before : you don't have to initialize user to null unless you want to be explicit, no?

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedUserDetailsService.java
```diff
@@ -30,27 +32,37 @@
...
     UserBuilder result = User.withUsername(email);
 
     result.password(user.getPasswordHash());
-    for (Role role : user.getRoles()) {
-      result.roles(role.getName());
+    String[] roles = new String[user.getRoles().size()];
+    for (int index = 0; index < user.getRoles().size(); index++) {
+      roles[index] = user.getRoles().get(index).getName();
     }
+    result.roles(roles);
```
BRUCELLA2: Couldn't something with forEach + lambda expression of new java versions be implemented?
Something like this (I don't know if it works, I haven't tested) :
    List<String> roles = new ArrayList<>();
    user.getRoles().forEach(role -> roles.add(role.getName()));
    result.roles(String.valueOf(roles));

comixed-rest-api/src/main/java/org/comixedproject/controller/tasks/TaskController.java
```diff
@@ -56,26 +57,29 @@
...
     log.debug("Getting all task audit log entries after: {}", cutoff);
 
     try {
       final List<TaskAuditLogEntry> entries = this.taskService.getAuditLogEntriesAfter(cutoff);
-      result.setResult(new GetTaskAuditLogResponse());
-      result.getResult().setEntries(entries);
-      result.getResult().setLatest(entries.get(entries.size() - 1).getStartTime());
-      result.setSuccess(true);
+      response.setResult(new GetTaskAuditLogResponse());
+      response.getResult().setEntries(entries);
```
BRUCELLA2, line 71: set the result then get the result to set entries seems strange for me. 
Wouldn't it be more understandable to do it like this:
    try {
      final List<TaskAuditLogEntry> entries = this.taskService.getAuditLogEntriesAfter(cutoff);
      GetTaskAuditLogResponse taskAuditLogResponse = new GetTaskAuditLogResponse();
      taskAuditLogResponse.setEntries(entries);
      if (!entries.isEmpty()) {
        taskAuditLogResponse.setLatest(entries.get(entries.size() - 1).getStartTime());
      }
      response.setResult(taskAuditLogResponse);
      response.setSuccess(true);

comixed-rest-api/src/main/java/org/comixedproject/authentication/AuthenticationConstants.java
```diff
@@ -16,7 +16,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses>
  */
 
-package org.comixedproject.web.authentication;
+package org.comixedproject.authentication;
 
 public interface AuthenticationConstants {
```
BRUCELLA2: According to sonar, it would be better to make a final class rather than an interface.

BRUCELLA2 (changes requested):
I don't have a lot of knowledge in the field but it seems to me that the md5 is not very secure and it might be better to use BCryptPasswordEncoder provided by Spring Security to encode passwords.

There is few sonar issues to fix too.

comixed-model/src/main/java/org/comixedproject/model/auditlog/RestAuditLogEntry.java
mcpierce: DOH! Fixed.

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationProvider.java
mcpierce: Java doesn't auto-initialize local variables to null, only instance variables.

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedUserDetailsService.java
mcpierce: See previous reply.

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedUserDetailsService.java
mcpierce: The following call requires an array of String rather than a list. I could have done the above, but would have then it'd still have an extra call to create an array from the List. I think with the extra work it was just simpler to create the array directly.

comixed-rest-api/src/main/java/org/comixedproject/authentication/AuthenticationConstants.java
mcpierce: Yeah, I had originally thought any entity who used it would implement the interface to have the values handy. I want to do a better job of this but not right now. ;)

BRUCELLA2:
I have merged #479 before this PR. There are now some conflicts to be solved...

mcpierce:
> I don't have a lot of knowledge in the field but it seems to me that the md5 is not very secure and it might be better to use BCryptPasswordEncoder provided by Spring Security to encode passwords.
> 
> There is few sonar issues to fix too.

I agreed about MD5 vs BCrypt. Can you open that suggestion in a separate ticket? This story was solely meant to provide the REST audit logging.

BRUCELLA2 (approved):

Issue #474, opened 2020-08-18
Use aspects to capture runtime statistics pm REST APIs


mcpierce:
Statistics to be captured include:
 * Remote IP address
 * User (if not anonymouse)
 * Runtime of the request.
 * Success of the request.
 * If an exception was caught.

PR #478, opened 2020-08-21
Make the scraping and build details APIs auditable

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Also updated the Angular state wiring for both feature sets.

**NOTE:** The PR depends on code from #474, which needs to be reviewed and merged first.


BRUCELLA2 (approved):

BRUCELLA2 (changes requested):
sonar reports 2 problems that can be corrected

mcpierce:
> sonar reports 2 problems that can be corrected

Just pushed and update that fixes both issues.

BRUCELLA2 (approved):

Issue #468, opened 2020-08-14
The task audit log page refresh ends up showing errors


mcpierce:
While on the task audit log page, new entries should be loaded without needing to reload the page.

mcpierce:
Forgot this was already implemented.

PR #486, opened 2020-08-23
Fixed loading task audit log entries [#468]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
The issue was mainly with the return value when no entries were found. The API was returning an empty latest value, which caused the frontend to exclude that parameter on subsequent requests. It now returns the current timestamp as the latest when no records are found.

### DIFF
--- a/comixed-frontend/src/app/backend-status/actions/load-task-audit-log.actions.ts
+++ b/comixed-frontend/src/app/backend-status/actions/load-task-audit-log.actions.ts
@@ -20,10 +20,17 @@ import { createAction, props } from '@ngrx/store';
 import { TaskAuditLogEntry } from 'app/backend-status/models/task-audit-log-entry';
 
 /**
+ * Starts the process of loading task audit log entries.
+ */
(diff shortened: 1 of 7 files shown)


BRUCELLA2 (approved):
PR approved but there is one conflict : 
comixed-rest-api/src/main/java/org/comixedproject/controller/tasks/TaskController.java

mcpierce:
> PR approved but there is one conflict :
> comixed-rest-api/src/main/java/org/comixedproject/controller/tasks/TaskController.java

Very strange, since the local copy of both branches had no conflicts and were both up to date. I just force pushed what I had to make sure the branch is good.

BRUCELLA2:
> > PR approved but there is one conflict :
> > comixed-rest-api/src/main/java/org/comixedproject/controller/tasks/TaskController.java
> 
> Very strange, since the local copy of both branches had no conflicts and were both up to date. I just force pushed what I had to make sure the branch is good.

Yes it's very strange.

Issue #465, opened 2020-08-10
Standardize the REST API response bodies.


mcpierce:
Remove the ApiResponse class as an interesting experiment that ultimately wasn't useful.

PR #490, opened 2020-08-27
Changed comic file import APIs to be auditable [#465]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
 * Updated the REST APIs for finding and importing comic files to be auditable. 
 * Updated the aspects to also log APIs that don't returning an ApiResponse object.
 * Replaced the frontend comic import APIs to use selectors.


comixed-frontend/src/app/comic-import/comic-import.fixtures.ts
```diff
@@ -13,24 +13,12 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses>
+ * along with this program. If not, see <http:/www.gnu.org/licenses>
```
BRUCELLA2: a "/" has disappeared.

comixed-frontend/src/app/comic-import/effects/selected-comic-files.effects.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import * as SelectedComicFilesActions from '../actions/selected-comic-files.actions';
+
+@Injectable()
+export class SelectedComicFilesEffects {
+  constructor(private actions$: Actions) {}
+
+  loadSelectedComicFiless$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(SelectedComicFilesActions.clearComicFileSelections),
```
BRUCELLA2: Couldn't we import only the clearCOmicFIleSelections action?
import { clearComicFileSelections } from '../actions/selected-comic-files.actions';
And use it : 
ofType(clearComicFileSelections),

comixed-frontend/src/app/comic-import/index.ts
```diff
@@ -30,14 +38,16 @@ interface RouterStateUrl {
...
 export type State = AppState;
 
 export const reducers: ActionReducerMap<AppState> = {
   router: fromRouter.routerReducer,
-  comicImport: fromImporting.reducer
+  [FIND_COMIC_FILES_FEATURE_KEY]: fromFindComicFiles.reducer,
+  [SELECTED_COMIC_FILES_FEATURE_KEY]: fromSelectedComicFiles.reducer
 };
 
 export const metaReducers: MetaReducer<AppState>[] = !environment.production
```
BRUCELLA2, line 53: This conditional operation returns the same value whether the condition is "true" or "false" => [] 
Couldn't we simplify this?

BRUCELLA2 (changes requested):

comixed-frontend/src/app/comic-import/comic-import.fixtures.ts
mcpierce: Restored it.

comixed-frontend/src/app/comic-import/index.ts
mcpierce: We could. But, like the empty SCSS files, I've left them in since we will take advantage of MetaReducers in future once the API is fully matured.

comixed-frontend/src/app/comic-import/effects/selected-comic-files.effects.ts
mcpierce: Yes. I've fixed it and updated the PR.

BRUCELLA2 (approved):

PR #492, opened 2020-08-28
Changed where and when the persisted task is deleted [#487]

### LINKED ISSUE #487: The persisted Task entry should not be removed until the task completes
**Describe the bug**
If one starts importing comics and then restarts the server, any comic that was in the midst of being processed is left in an in-between state; ie, it has no ComicFileDetails associated with it, but there are no tasks enqueued to process it. This is because the task running system deletes the record from the tasks table before the task itself completes running. Instead, the entire thing should be handled in a single transaction: 

1. Load the **tasks** entry.
1. Rehydrate the task.
1. Execute the task.
1. Delete the entry from the **tasks** table.

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Previously, the persisted task was deleted from the database when it was
decoded and before it was enqueued.

Now the persisted task, along with the executable one, is given to the
TaskManager. When the executable task completes then the persisted one
is removed from the database.

### DIFF
--- a/comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicWorkerTask.java
+++ b/comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicWorkerTask.java
@@ -22,6 +22,8 @@ import java.io.File;
 import java.io.IOException;
 import java.text.MessageFormat;
 import java.util.Date;
+import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
 import org.apache.commons.io.FileUtils;
 import org.comixedproject.model.comic.Comic;
@@ -44,12 +46,8 @@ public class DeleteComicWorkerTask extends AbstractWorkerTask implements WorkerT
   @Autowired private ComicService comicService;
   @Autowired private ReadingListService readingListService;
 
-  private boolean deleteFile;
-  private Comic comic;
-
-  public void setDeleteFile(boolean deleteFile) {
-    this.deleteFile = deleteFile;
-  }
+  @Getter @Setter private Boolean deleteFile;
+  @Getter @Setter private Comic comic;
 
   @Override
   @Transactional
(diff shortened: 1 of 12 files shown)


comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicWorkerTask.java
```diff
@@ -44,12 +46,8 @@
   @Autowired private ComicService comicService;
   @Autowired private ReadingListService readingListService;
 
-  private boolean deleteFile;
-  private Comic comic;
-
-  public void setDeleteFile(boolean deleteFile) {
-    this.deleteFile = deleteFile;
-  }
+  @Getter @Setter private Boolean deleteFile;
```
BRUCELLA2, line 49: Is the change from boolean to boolean intentional?

comixed-tasks/src/main/java/org/comixedproject/task/runner/TaskManager.java
```diff
@@ -43,26 +44,45 @@
...
           }
         });
   }
 
+  /**
+   * Enqueue and execute the given task.
+   *
+   * @param task the task
+   */
+  public void runTask(final WorkerTask task) {
```
BRUCELLA2: workerTask serait mieux que task pour éviter les confusion (entre la classe WorkerTask et la class Task).

BRUCELLA2 (changes requested):

comixed-tasks/src/main/java/org/comixedproject/task/model/DeleteComicWorkerTask.java
mcpierce: Yes. Since it's coming out of the database, I'm using the wrapper class to defensively program against the value being null.

comixed-tasks/src/main/java/org/comixedproject/task/runner/TaskManager.java
mcpierce: D'accord. Je l'ai renommé.

BRUCELLA2: ho sorry ... I wrote in French !!!! :D 
For non French speaking people :
workerTask would be better than task to avoid confusion (between the WorkerTask class and the Task class).

BRUCELLA2 (approved):

PR #497, opened 2020-09-06
Added scrubbing of filename components when moving comics [#491]

### LINKED ISSUE #491: Handle special characters when consolidating the library.
Whenever a filename is created based on either incoming renaming rules or from metadata for a comic, the final name should be scrubbed to ensure it contains no characters or substrings that could cause problems.

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Replace invalid characters in the renaming rule and also in a comic's metadata when creating a new filename.

### DIFF
--- a/comixed-tasks/src/main/java/org/comixedproject/task/model/MoveComicWorkerTask.java
+++ b/comixed-tasks/src/main/java/org/comixedproject/task/model/MoveComicWorkerTask.java
@@ -45,8 +45,11 @@ import org.springframework.util.StringUtils;
 @Scope(value = ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 @Log4j2
 public class MoveComicWorkerTask extends AbstractWorkerTask {
-  static final SimpleDateFormat dateFormat = new SimpleDateFormat("MMM yyyy");
+  private static final String FORBIDDING_RULE_CHARACTERS = "[:\\\\*?|<>]";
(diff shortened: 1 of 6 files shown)


comixed-tasks/src/main/java/org/comixedproject/task/model/MoveComicWorkerTask.java
```diff
@@ -120,26 +120,30 @@ String getRelativeComicFilename() {
       return FilenameUtils.getBaseName(this.comic.getFilename());
     }
 
-    log.debug("Generating relative filename based on renaming rule: {}", this.renamingRule);
+    log.debug("Scrubbing renaming rule: {}", this.renamingRule);
+    final String rule = this.scrub(this.renamingRule, "[:\\\\*?|<>]");
```
BRUCELLA2: Because these are characters that are globally forbidden and because this list of characters is used in several places, wouldn't it be better to define it as a constant?
We can use this.scrub(this.renamingRule) too, no ?

BRUCELLA2 (changes requested):

comixed-tasks/src/main/java/org/comixedproject/task/model/MoveComicWorkerTask.java
mcpierce: Sure, that works! Fixed it. I also added scrubbing the rules as they're entered on the browser to avoid submitted invalid characters as well.

BRUCELLA2 (approved):

Issue #371, opened 2020-06-27
Provide Portuguese language translation 


mcpierce:
Provide support for viewing the application in Portuguese.

mcpierce:
@jmcrfp I've merged the initial part of this so you can build on it with the translation work.

PR #499, opened 2020-09-08
Support for portuguese [#371]

jmcrfp:
## Status
**READY**

## Migrations
NO

## Description
Support for portuguese

### DIFF
--- a/comixed-frontend/src/assets/i18n/pt/app.json
+++ b/comixed-frontend/src/assets/i18n/pt/app.json
@@ -1,133 +1,133 @@
 {
   "menu": {
     "home": {
-      "root": "Home"
+      "root": "Início"
     },
     "library": {
-      "root": "Library",
+      "root": "Biblioteca",
       "comics": "Comics",
-      "import": "Import",
-      "duplicate-pages": "Duplicate Pages",
-      "missing-comics": "Missing Comics"
+      "import": "Importar",
+      "duplicate-pages": "Páginas duplicadas",
+      "missing-comics": "Comics em falta"
     },
     "admin": {
-      "root": "Admin",
-      "users": "Users",
-      "library": "Library"
+      "root": "Administrador",
+      "users": "Utilizadores",
+      "library": "Biblioteca"
     },
     "help": {
-      "root": "Help",
-      "build-details": "Build Details"
+      "root": "Ajuda",
+      "build-details": "Informação da compilação"
     }
   },
   "app-component": {
     "copyright": "<b>Copyright &copy; {years}, The ComiXed Project Team</b>",
-    "library-counts": "<b>Total:</b> {comic_count} <b>Read:</b> {read_count} <b>Processing:</b> {import_count}"
+    "library-counts": "<b>Total:</b> {comic_count} <b>Lidos:</b> {read_count} <b>Processando:</b> {import_count}"
   },
   "language": {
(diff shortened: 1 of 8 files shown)


mcpierce:
Will you be pushing a follow up PR for the translations? I see that the backend-status.json file isn't translated in this PR.

comixed-frontend/src/assets/i18n/pt/library.json
```diff
@@ -169,378 +169,378 @@
...
-    "title": "ComiXed: Story Arcs ({count})"
+    "title": "ComiXed: Arcos da historias ({count})"
   },
   "story-arc-details-page": {
-    "title": "ComiXed: Story Arc - {name} ({count, plural, =1{One Comic} other{# Comics}})"
+    "title": "ComiXed: Arco da historia - {name} ({count, plural, =1{Um comic} other{# comics}})"
   },
   "duplicates-page": {
-    "title": "{pageCount} Duplicate {pageCount, plural, =1{Page} other{Pages}}",
+    "title": "{pageCount} {pageCount, plural, =1{Pagina} other{Paginas} duplicadas}",
```
mcpierce: The final curly brace on this entry is out of place. It should come after Paginas as follows:
```
"title": "{pageCount} {pageCount, plural, =1{Pagina} other{Paginas}} duplicadas",
```

comixed-frontend/src/assets/i18n/pt/app.json
```diff
@@ -316,136 +316,136 @@
...
     "errors": {
-      "form-field-required": "{name} is required."
+      "form-field-required": "{name} é necessário."
     },
     "tooltip": {
-      "unprocessed-comic": "This comic is scheduled for processing.",
-      "deleted-comic": "This comic is marked for deletion."
+      "unprocessed-comic": "Este comic está agendado para ser processado.",
+      "deleted-comic": "Este comic está selecionado para ser apagado."
     }
```
mcpierce, line 437: I found a missing translation from #376 that wasn't added to this file. Can you add the following to the block below tooltip and translate it, please?

```
    "confirmation": {
      "logout": {
        "header": "Logout",
        "message": "Are you sure you want to log out?"
      }
    }
```

mcpierce (changes requested):
All in all looking good. Just one small fix, an addition and one other file to translate and we can merge this.

comixed-frontend/src/assets/i18n/pt/library.json
jmcrfp: Updated

comixed-frontend/src/assets/i18n/pt/app.json
jmcrfp: Added to the file

jmcrfp:
> Will you be pushing a follow up PR for the translations? I see that the backend-status.json file isn't translated in this PR.

The file was translated. I thought I translated everything. Did I miss something?

mcpierce:
> > Will you be pushing a follow up PR for the translations? I see that the backend-status.json file isn't translated in this PR.
> 
> The file was translated. I thought I translated everything. Did I miss something?

Hrm, I had thought it was still english. But it seems i was mistaken. The issue is that, in the component itself, the i18n entry name was changed so it differs from the files now.

mcpierce (approved):

Issue #332, opened 2020-06-10
Provide Spanish language translations


mcpierce:
Provide a translation of the application's texts in Spanish.

PR #507, opened 2020-09-16
Updated Spanish translation for 0.7 [#332]

theotocopulitos:
## Status
READY

## Migrations
NO

## Description
Updated Spanish translation for 0.7 [#332]

### DIFF
--- a/comixed-frontend/src/assets/i18n/es/backend-status.json
+++ b/comixed-frontend/src/assets/i18n/es/backend-status.json
@@ -2,24 +2,24 @@
   "backend-status": {
     "effects": {
       "get-rest-audit-log-entries": {
-        "error-detail": "Failed to load audit log entries..."
+        "error-detail": "Error al cargar las entradas en el log de auditoría de web..."
       }
     },
     "rest-audit-log": {
       "page-title": "Web Logs",
       "entries": {
         "header": {
-          "remote-ip": "Originating IP",
-          "url": "Endpoint URL",
-          "email": "Account",
-          "start-time": "Started",
-          "end-time": "Completed",
-          "bytes-received": "Bytes In",
-          "bytes-returned": "Bytes Out",
-          "success": "Success",
-          "request-content": "Request Content",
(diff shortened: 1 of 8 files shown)


comixed-frontend/src/assets/i18n/es/backend-status.json
```diff
@@ -30,7 +30,7 @@
         "detail": "Error al recabar los datos de compilación..."
       }
     },
-    "page-title": "Build Details",
+    "page-title": "Detalles de la compilación",
     "table": {
```
mcpierce, line 34: Can you translate these entries as well?

mcpierce (changes requested):
I saw a few entries that didn't get translated. Could you review the files and fix those entries, please?

For example, in library.json the clear-image-cache entries are still in english.

theotocopulitos:
Sorry :( ... not sure how I missed that...

But, regarding these lines:

    "page-title": "Detalles de la compilación",
    "table": {
      "branch": "Source Branch",
      "build-host": "Build Hostname",
      "build-time": "Built On",
      "build-version": "Version",
      "commit-user": "Committer Name",
      "commit-id": "Commit Hash",
      "commit-message": "Commit Message",
      "commit-time": "Time Of Last Commit",
      "dirty": "Built From Uncommitted Code?",
      "remote-origin-url": "Remote Origin",
      "jdbc-url": "JDBC Connection URL"
    }

I'd rather not translate them... this info relates to git, and there is not
a universal translation of those concepts. Anyone looking at these uses
them in English, I am guessing. Should you prefer me to translate that, I
will, but honestly, I think we should not.

mcpierce:
Fair enough. Do other applications translate similar entries? I believe @BRUCELLA2 and @jmcrfp translated them, but you make a good point about them not being something to translate. Let's leave them untranslated here and we can discuss it in the dev list.

theotocopulitos:
I **think** I was able to squash all changes into a single commit!

mcpierce (approved):

Issue #483, opened 2020-08-22
Admins can clear the REST API audit log.


mcpierce:
As an administrator, I would like to be able to clear the entries from the audit log.

PR #533, opened 2020-10-26
Added clear REST API audit log [#483]

BRUCELLA2:
## Status
**READY**

## Migrations
NO

## Description
* Added a button to clear the REST API audit log (only for Admin)

### DIFF
--- a/comixed-frontend/src/app/backend-status/effects/clear-rest-audit-log.effects.ts
+++ b/comixed-frontend/src/app/backend-status/effects/clear-rest-audit-log.effects.ts
@@ -1,6 +1,6 @@
 /*
  * ComiXed - A digital comic book library management application.
- * Copyright (C) 2019, The ComiXed Project
+ * Copyright (C) 2020, The ComiXed Project
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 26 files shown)


BRUCELLA2:
I forgot the effects tests. I will add them.

comixed-frontend/src/app/backend-status/effects/clear-rest-audit-log.effects.ts
```diff
@@ -0,0 +1,94 @@
...
+                  )
+                )
+              : this.alertService.error(
+                  this.translateService.instant(
+                    'rest.clear-rest-audit-log.effects.clear-audit-log.error.detail'
+                  )
+                )
+          ),
+          map((response: ApiResponse<void>) =>
+            response.success ? restAuditLogCleared() : clearRestAuditLogFailed()
```
mcpierce: Would it be possible to have this return a mergeMap that includes the success action as well as an action to clear the loaded audit log?

mcpierce (changes requested):
One question for the initial review.

BRUCELLA2:
My pc is working properly again! So I made some modifications taking into account your remarks. I added the refresh of the rest log list after the clean. The code works but I'm not sure it's super clean because I'm not too familiar with actions, effects, reducers and selectors...

mcpierce (approved):
All looks good! Thanks for this!

Issue #534, opened 2020-11-05
Removing comics from a reading list - action confirmation toaster showing placeholder text.


bareheiny:
As per title.  I removed a number of comics from a reading list, and the action confirmation toaster showed place holder text.  Unfortunately I was not quick enough to grab a screen capture.

Build details:

Source Branch | f57c46da8896b205a7aa1f45fb160a05be7009c7
-- | --
Build Hostname | fv-az50
Built On | Oct 2, 2020, 5:16:58 PM
Version | 0.7.0-rc4
Commit Hash | f57c46da8896b205a7aa1f45fb160a05be7009c7
Committer Name | Darryl L. Pierce ()
Time Of Last Commit | Oct 2, 2020, 5:10:13 PM
Commit Message | Changed the release version to 0.7.0-rc4 [#496]
Built From Uncommitted Code? | true
Remote Origin |  
JDBC Connection URL | jdbc:h2:D:\Comics\.comixedDB\comixed

mcpierce:
@BRUCELLA2 @theotocopulitos  @jmcrfp  I added you as additional assignees. Could you each please provide translations for the new entries? Thanks.

PR #536, opened 2020-11-09
Added missing translations for removing comics from a reading list [#534]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Added missing translation to all language files. Non-english entries still need to be translated.


theotocopulitos:
Darryl,

Familia health issues keep me busy and in and out of the hospital as a companion, so I will not be able to tackle this properly. 

If you could send me just the sentences that need translation, I could do that and then you put them back where they belong.

Otherwise, I have no problem if you find another enthusiast willing to translate.

It seems I will be involved in this situation for a while...

Arturo

Issue #328, opened 2020-06-09
Migrate to Angular Material for the user interface


mcpierce:
Use Material for the user interface.

mcpierce:
This was fixed in #537.

PR #537, opened 2020-11-12
Added the new comixed-web submodule to replace the UX [#328]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Created a new submodule, comixed-web, that provides an Angular app based on Angular Material for the user interface. Add the initial features to allow a user to login and logout, a confirmation dialog and an alert service, and a simple feature set based on NgRx.


comixed-rest-api/src/test/java/org/comixedproject/controller/user/UserControllerTest.java
```diff
@@ -138,7 +139,10 @@ public void testSetUserPropertyForInvalidUser() throws ComiXedUserException {
 
   @Test
   public void testGetCurrentUserWhenNotAuthented() throws ComiXedUserException {
```
BRUCELLA2: it's a detail but it seems to me that there is a mistake in the name of the method (but I can be wrong, my English is not very good).

comixed-web/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+        <version>1.7.5</version>
+        <executions>
+          <execution>
+            <id>Install node and yarn</id>
+            <goals>
+              <goal>install-node-and-yarn</goal>
+            </goals>
+            <phase>initialize</phase>
+            <configuration>
+              <nodeVersion>v14.7.0</nodeVersion>
```
BRUCELLA2: lastest version for node is v14.15.0. Do we need to update it ?

BRUCELLA2 (commented):
It's ok for me, juste 2 questions / comments.

comixed-rest-api/src/test/java/org/comixedproject/controller/user/UserControllerTest.java
mcpierce: Good catch. Fixed it.

comixed-web/pom.xml
mcpierce: I'm good with updating things to the latest. Done!

BRUCELLA2 (approved):

Issue #539, opened 2020-11-14
Reimplement the comic import interface.


mcpierce:
This issue depends on #283.

Re-implement the user interface for importing comics using Angular Material.

PR #540, opened 2020-11-16
Reimplemented the import comics page [#539]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
Whole new implementation of the comic import page using Angular Material.


comixed-web/src/app/app.component.spec.ts
```diff
@@ -23,20 +23,33 @@ import { LoggerModule } from '@angular-ru/logger';
 import { MockStore, provideMockStore } from '@ngrx/store/testing';
 import {
   initialState as initialUserState,
-  USER_FEATURE_KEY
+  USER_FEATURE_KEY,
```
BRUCELLA2: do you change your prettier version ? With Prettier 1.19.1 , comma is removed but with 2.0 and better trailing comma is not removed by default.

comixed-web/src/app/app.component.ts
```diff
@@ -22,21 +22,32 @@ import { LoggerService } from '@angular-ru/logger';
...
 import { User } from '@app/user/models/user';
 import { loadCurrentUser } from '@app/user/actions/user.actions';
+import { selectBusyState } from '@app/core/selectors/busy.selectors';
+import { TranslateService } from '@ngx-translate/core';
 
 @Component({
   selector: 'cx-root',
   templateUrl: './app.component.html',
-  styleUrls: ['./app.component.scss']
+  styleUrls: ['./app.component.scss'],
```
BRUCELLA2: do you change your prettier version ? With Prettier 1.19.1 , comma is removed but with 2.0 and better trailing comma is not removed by default.

comixed-web/src/app/app.component.ts
```diff
@@ -22,21 +22,32 @@ import { LoggerService } from '@angular-ru/logger';
...
-  constructor(private logger: LoggerService, private store: Store<any>) {
+  constructor(
+    private logger: LoggerService,
+    private translateService: TranslateService,
+    private store: Store<any>
+  ) {
+    this.translateService.use('en');
     this.logger.trace('Subscribing to user changes');
-    this.store.select(selectUser).subscribe(user => {
+    this.store.select(selectUser).subscribe((user) => {
```
BRUCELLA2: I think it's Prettier 2.0 ? (arrowParens set to always by default)

BRUCELLA2 (commented):
Just one question about your Prettier version.
And an other thing, Sonar does not take into account the coverage of the comixed-web module.

comixed-web/src/app/app.component.spec.ts
mcpierce: I still had Prettier 1.18.2, but upgraded now to 1.19.1 as latest. That removed the trailing commas, so I'll push an update on this.

comixed-web/src/app/app.component.ts
mcpierce: Like the other spots, this is formatted differently with with 1.19.1.

mcpierce:
> Just one question about your Prettier version.
> And an other thing, Sonar does not take into account the coverage of the comixed-web module.

I'm not sure why it's ignoring the comixed-web project. I dug up your email from before when we fixed the SonarCloud reporting and applied those changes here. So let's hope it fixes things in this regard.

BRUCELLA2 (approved):

Issue #548, opened 2020-12-05
Library administrators can edit the details for a single comic.


mcpierce:
As a library administrator, I want to be able to change the details for a single comic from the comic details page. Fields are changed and submitted as a single set of changes to the database.

The details will be divided into two sections:
 * Top: the minimum fields needed to scrape a comic, and
 * Bottom: the remaining details for a comic.

The Scrape button will be enabled if the minimum fields are filled in for scraping a comic (already done). The save button will be enabled at all times since there is no minimum required set of values that defines a comic.

PR #579, opened 2020-12-30
Added the ability to change comic details for a single comic [#548]

BRUCELLA2:
## Status
**READY**

## Description
Added the ability to change comic details for a single comic for admin

### DIFF
--- a/comixed-web/src/app/library/actions/comic-details.actions.ts
+++ b/comixed-web/src/app/library/actions/comic-details.actions.ts
@@ -1,0 +1,41 @@
+/*
+ * ComiXed - A digital comic book library management application.
+ * Copyright (C) 2020, The ComiXed Project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses>
+ */
+
+import { createAction, props } from '@ngrx/store';
+import { Comic } from '@app/library';
+/*
+export class SaveComicDetails implements Action {
+  readonly type = '[COMIC] Save changes to a comic';
+
+  constructor(public payload: { comic: Comic }) {}
+}
+
+ */
+export const saveComicDetails = createAction(
+  '[COMIC] Save changes to a comic',
+  props<{ comic: Comic }>()
+);
+
+export const comicSaved = createAction(
+  '[COMIC] Changes to a comic were saved',
+  props<{ comic: Comic }>()
+);
+
+export const saveComicDetailsFailed = createAction(
+  '[COMIC] Failed to save changes to a comic'
+);
--- a/comixed-web/src/app/library/components/comic-edit/comic-edit.component.html
+++ b/comixed-web/src/app/library/components/comic-edit/comic-edit.component.html
@@ -5,6 +5,7 @@
     mat-flat-button
     color="primary"
     [disabled]="!comicForm.valid"
+    (click)='onSaveDetails()'
   >
     <mat-icon>save</mat-icon>
     <mat-label>{{ "button.save" | translate }}</mat-label>
--- a/comixed-web/src/app/library/components/comic-edit/comic-edit.component.spec.ts
+++ b/comixed-web/src/app/library/components/comic-edit/comic-edit.component.spec.ts
@@ -38,6 +38,7 @@ import {
   API_KEY_PREFERENCE,
   MAXIMUM_RECORDS_PREFERENCE
 } from '@app/library/library.constants';
+import { saveComicDetails } from '@app/library/actions/comic-details.actions';
 
 describe('ComicEditComponent', () => {
   const COMIC = COMIC_2;
@@ -96,6 +97,34 @@ describe('ComicEditComponent', () => {
 
     it('updates the comic reference', () => {
       expect(component.comic).toEqual(COMIC);
+    });
+  });
+
+  describe('save comic details', () => {
+    beforeEach(() => {
+      spyOn(
+        confirmationService,
+        'confirm'
+      ).and.callFake((confirmation: Confirmation) => confirmation.confirm());
+      component.onSaveDetails();
+    });
+
+    it('confirms with the user', () => {
+      expect(confirmationService.confirm).toHaveBeenCalled();
+    });
+
+    it('fire an action', () => {
+      expect(store.dispatch).toHaveBeenCalledWith(
+        saveComicDetails({
+          comic: {
+            ...COMIC_2,
+            publisher: COMIC.publisher,
+            series: COMIC.series,
+            volume: COMIC.volume,
+            issueNumber: COMIC.issueNumber
+          }
+        })
+      );
     });
   });
 
--- a/comixed-web/src/app/library/components/comic-edit/comic-edit.component.ts
+++ b/comixed-web/src/app/library/components/comic-edit/comic-edit.component.ts
@@ -30,6 +30,7 @@ import {
   MAXIMUM_RECORDS_PREFERENCE,
   SKIP_CACHE_PREFERENCE
 } from '@app/library/library.constants';
+import { saveComicDetails } from '@app/library/actions/comic-details.actions';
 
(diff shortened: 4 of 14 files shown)


comixed-web/src/app/library/actions/comic-details.actions.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Comic } from '@app/library';
+/*
+export class SaveComicDetails implements Action {
+  readonly type = '[COMIC] Save changes to a comic';
+
+  constructor(public payload: { comic: Comic }) {}
+}
+
+ */
+export const saveComicDetails = createAction(
```
mcpierce, line 29: Can we rename this and the failure action to "saveComic" and "saveComicFailed", to keep the names consistent with the success action?

comixed-web/src/app/library/components/comic-edit/comic-edit.component.html
```diff
@@ -5,6 +5,7 @@
     mat-flat-button
     color="primary"
     [disabled]="!comicForm.valid"
+    (click)='onSaveDetails()'
```
mcpierce, line 8: Small nit: this is using single instead of double quotes.

There's a yarn target (yarn code-cleanup) that you can run before checking in Angular code that will go through and enforce coding stands in the same way we're using Cosium to enforce conventions on the Java code. Please run that on comixed-web and push the updated commit.

comixed-web/src/app/library/components/comic-edit/comic-edit.component.spec.ts
```diff
@@ -99,6 +100,34 @@ describe('ComicEditComponent', () => {
     });
   });
 
+  describe('save comic details', () => {
```
mcpierce, line 103: Can you change the text here to be "saving comic details"?

comixed-web/src/app/library/components/comic-edit/comic-edit.component.ts
```diff
@@ -100,6 +101,29 @@ export class ComicEditComponent implements OnInit {
...
+            comic: {
+              ...this._comic,
+              publisher: this.comicForm.controls.publisher.value,
+              series: this.comicForm.controls.series.value,
+              volume: this.comicForm.controls.volume.value,
+              issueNumber: this.comicForm.controls.issueNumber.value
+            }
+          })
+        );
+        this.comicForm.markAsPristine();
```
mcpierce, line 122: Let's remove this line here. The form should only be marked as pristine when the updated comic is received from the backend and the form is reloaded with the values received.

comixed-web/src/app/library/effects/comic.effects.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    spyOn(alertService, 'error');
+  });
+
+  it('should be created', () => {
+    expect(effects).toBeTruthy();
+  });
+
+  describe('when saving comic detail', () => {
+    it('fire an action on success', () => {
+      const serviceResponse = COMIC_1 as Comic;
```
mcpierce, line 85: The "as Comic" piece here isn't necessary, since COMIC_1 is already an instance of the Comic interface.

That said, can we instead declare a constant named COMIC at the top of this testing suite and use that in all the tests instead? That way we can re-use it for other tests that'll be written in future.

comixed-web/src/app/library/effects/comic.effects.ts
```diff
@@ -0,0 +1,80 @@
...
+    private comicService: ComicDetailsService,
+    private alertService: AlertService,
+    private translateService: TranslateService,
+    private actions$: Actions
+  ) {}
+
+  saveComicDetail$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(saveComicDetails),
+      tap(() => this.logger.debug('effect: save comic details')),
```
mcpierce, line 47: For consistency, please make the debug text start with 'Effect:'.

comixed-web/src/app/library/effects/comic.effects.ts
```diff
@@ -0,0 +1,80 @@
...
+          tap(() =>
+            this.alertService.info(
+              this.translateService.instant(
+                'comic.comic-effects.save-comic.success.detail'
+              )
+            )
+          ),
+          map((response: Comic) => comicSaved({ comic: response })),
+          catchError(error => {
+            this.logger.error('service failure save comic:', error);
```
mcpierce, line 59: For service and general failure, the convention is to just use "Service failure:" and "General failure:" in the logger output. Please update this line and the general failure below to match that.

comixed-web/src/app/library/effects/comic.effects.ts
```diff
@@ -0,0 +1,80 @@
...
+            );
+            return of(saveComicDetailsFailed);
+          })
+        )
+      ),
+      catchError(error => {
+        this.logger.error('geneservice failure save comic:', error);
+        this.alertService.error(
+          this.translateService.instant(
+            'general-message.error.general-service-failure'
```
mcpierce, line 73: In the translation files the key for this message is "app.general-effect-failure". Please update this line.

comixed-web/src/app/library/services/comic-details.service.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  it('can save details comic', () => {
+    const serviceResponse = new HttpResponse({ status: 200 });
+    service
+      .saveComic(COMIC_1)
```
mcpierce, line 51: Please set a test suite-level constant named COMIC and use that here.

comixed-web/src/app/library/services/comic-details.service.ts
```diff
@@ -0,0 +1,41 @@
...
+})
+export class ComicDetailsService {
+  constructor(private logger: LoggerService, private http: HttpClient) {}
+
+  /**
+   * Save the comic with modified details
+   *
+   * @param comic with modified details
+   */
+  saveComic(comic: Comic): Observable<any> {
```
mcpierce, line 38: The service arguments should be wrapped in a single typed variable named "args" (see the other services for an example). Please change this to:
```
saveComic(args:{comic: Comic}): Obserable<any>
```

Also, we should keep things together. Right now the method to load a comic is in LibraryService. I don't think we should create a new service for saving comics. Can you move this API into the LibraryService class? It's okay to split out features based on functional behavior, but lets keep the services in as few classes as necessary.

comixed-web/src/app/library/services/comic-details.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    return this.http.put(interpolate(SAVE_COMIC_URL, { id: comic.id }), comic);
```
mcpierce, line 39: Please add a  logging line before the HTTP action to show ```args``` object. See the ```loadComic()``` method in LibraryService.

comixed-web/src/assets/i18n/en/library.json
```diff
@@ -102,6 +106,16 @@
       "download-comic": "Download this comic...",
       "go-to-next-comic": "Go to the next comic in this series...",
       "go-to-previous-comic": "Go to the previous comic in this series..."
+    },
+    "comic-effects": {
```
mcpierce, line 110: In the web project, I've been doing things differently than in the frontend project (learned some good lessons). For example, everything to do with scraping a comic is under the "scraping" group in the translations.

Since this is related to a single comic, can we move these translations to a group named "comic"? So it would be (without the comments, of course :smile:):

```
{
  "comic": { // the grouping 
    "save-comic": {  // the specific feature in that group
      "effect-success": "Changes successfully saved.",
      "effect-failure": "Failed to save changes."
    }
  }
```

Using "effect-success" and "effect-failure" keeps things simple when searching for an entry.

mcpierce (changes requested):
We'll want to leave the ticket open after the requested changes are added the PR merged. Ticket #548 was intended to allow the admin to edit everything about the comic; i.e., the scan type, publisher, sort name, etc. But we can discuss those changes in the dev group and do a follow-up PR to add those fields to the form.

Also, please go into the REST API and add the following to the ComicController.updateComic() endpoint:
```
 @PreAuthorize("hasRole('ADMIN')")
@AuditableEndpoint
```

comixed-web/src/app/library/actions/comic-details.actions.ts
```diff
@@ -0,0 +1,41 @@
+/*
```
mcpierce, line 1: I just realized there's no reducer on this PR. Looking it over, I think we should move all of this feature behavior into the existing Library feature where it's already managing the currently displayed comic. Then you would want to add a "saving" flag to the state and have the ComicDetails component set the busy indicator based on saving as well as loading.

mcpierce:
@BRUCELLA2  I'm closing this pull request as stale. If you could, please rebase it and target the new master branch for the next release.

Issue #578, opened 2020-12-30
A page displays the build details for the server.


mcpierce:
(this is a re-implementation)

Anonymous and authenticated users can view a page that displays the build details for the server.

PR #580, opened 2021-01-02
Added a feature to load the server build details [#578]

mcpierce:
## Status
**READY**

## Migrations
NO

## Description
 * Added the frontend feature to retrieve the details.
 * Refactored the REST APIs and repackaged the classes.
 * Added the page to display the build details.

### DIFF
--- a/comixed-web/src/app/effects/build-details.effects.ts
+++ b/comixed-web/src/app/effects/build-details.effects.ts
@@ -1,0 +1,72 @@
+/*
+ * ComiXed - A digital comic book library management application.
+ * Copyright (C) 2021, The ComiXed Project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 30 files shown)


comixed-web/src/app/effects/build-details.effects.ts
```diff
@@ -0,0 +1,72 @@
...
+    );
+  });
+
+  constructor(
+    private logger: LoggerService,
+    private actions$: Actions,
+    private buildDetailsService: BuildDetailsService,
+    private alertService: AlertService,
+    private translateService: TranslateService
+  ) {}
```
BRUCELLA2, line 71: Is it voluntary to place the builder at the end of the class?
In the other effects.ts files the constructor is placed at the beginning of the class. This is consistent with a field -> constructor -> method structure. If this is what you want, you will have to correct the other classes in the course of time.

BRUCELLA2 (changes requested):
it's ok just a question and we can merge

comixed-web/src/app/effects/build-details.effects.ts
mcpierce: I believe that's a side-effect of my IntelliJ setup and how it cleaned up the code. The constructor is the only pure method in the effect source module while the effect methods are generated by method calls to createEffect(). So IJ sorted those method calls above the constructor call. It's not a requirement for the code.

BRUCELLA2: Okay, that's good then.

BRUCELLA2 (approved):

Issue #615, opened 2021-03-01
Add STOMP messaging support.


mcpierce:
As a system, the frontend and backend exchange state changes via STOMP messages.

https://www.offerzen.com/blog/effective-communication-angular-ngrx-socketio

PR #641, opened 2021-03-09
Added STOMP messaging between the front and backends [#615]

mcpierce:
* Added the frontend messaging feature.
 * Added a queue for monitoring the number of active tasks.
 * Added a new feature to track server status.

## Status
READY

NO

## Before You Submit Your PR:

### DIFF
--- a/comixed-web/src/app/reducers/messaging.reducer.ts
+++ b/comixed-web/src/app/reducers/messaging.reducer.ts
@@ -1,0 +1,71 @@
+/*
+ * ComiXed - A digital comic book library management application.
+ * Copyright (C) 2021, The ComiXed Project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses>
+ */
+
+import { createReducer, on } from '@ngrx/store';
+import {
+  messagingStarted,
+  messagingStarting,
+  messagingStopped,
+  restartMessaging,
+  startMessaging,
+  startMessagingFailed,
+  stopMessaging,
+  stopMessagingFailed
+} from '@app/actions/messaging.actions';
+
+export const MESSAGING_FEATURE_KEY = 'messaging_state';
+
+export interface MessagingState {
+  startingMessaging: boolean;
+  stoppingMessaging: boolean;
+  messagingStarted: boolean;
+}
+
+export const initialState: MessagingState = {
+  startingMessaging: false,
+  stoppingMessaging: false,
+  messagingStarted: false
+};
+
+export const reducer = createReducer(
+  initialState,
+  on(startMessaging, state => ({
+    ...state,
+    startingMessaging: true,
+    messagingStarted: false
+  })),
+  on(messagingStarting, state => ({ ...state, startingMessaging: true })),
+  on(startMessagingFailed, state => ({ ...state, startingMessaging: false })),
+  on(messagingStarted, state => ({
+    ...state,
+    startingMessaging: false,
+    messagingStarted: true
+  })),
+  on(restartMessaging, state => ({
+    ...state,
+    startingMessaging: true,
+    messagingStarted: false
+  })),
+  on(stopMessaging, state => ({ ...state, stoppingMessaging: true })),
+  on(messagingStopped, state => ({
+    ...state,
+    stoppingMessaging: false,
+    messagingStarted: false
+  })),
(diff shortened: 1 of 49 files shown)


BRUCELLA2 (commented):
Some comments and questions

comixed-web/src/app/services/session.service.ts
```diff
@@ -23,12 +23,33 @@ import { SessionUpdateRequest } from '@app/models/net/session-update-request';
...
+    private http: HttpClient,
+    private taskCountService: TaskCountService
+  ) {}
+
+  /**
+   * Invoked to start subscriptions.
+   */
+  startSubscriptions(): void {
+    this.logger.debug('Starting session services');
+    this.taskCountService.start();
```
BRUCELLA2, line 43: Why does sessionService.startSubscriptions call taskCountService.start? Is this some kind of test to see if it works well? If not, what is the purpose of this task counter normally used for import count service? ?

comixed-web/src/app/reducers/messaging.reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export const reducer = createReducer(
+  initialState,
+  on(startMessaging, state => ({
+    ...state,
+    startingMessaging: true,
+    messagingStarted: false
+  })),
+  on(messagingStarting, state => ({ ...state, startingMessaging: true })),
```
BRUCELLA2, line 52: Shouldn't stoppingMessaging be set to false?

comixed-web/src/app/reducers/messaging.reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+    ...state,
+    startingMessaging: false,
+    messagingStarted: true
+  })),
+  on(restartMessaging, state => ({
+    ...state,
+    startingMessaging: false,
+    messagingStarted: false
+  })),
+  on(stopMessaging, state => ({ ...state, stoppingMessaging: true })),
```
BRUCELLA2, line 64: Shouldn't startingMessaging be set to false?

comixed-web/src/app/reducers/messaging.reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+    messagingStarted: false
+  })),
+  on(messagingStarting, state => ({ ...state, startingMessaging: true })),
+  on(startMessagingFailed, state => ({ ...state, startingMessaging: false })),
+  on(messagingStarted, state => ({
+    ...state,
+    startingMessaging: false,
+    messagingStarted: true
+  })),
+  on(restartMessaging, state => ({
```
BRUCELLA2, line 59: When restart messaging, there is a stop and start ? or just a start is messaging is already stopped ? Shouldn't startingMessaging be set to true ? and stoppingMessaging to false ? or true ? (i'm not sure)

comixed-web/src/app/selectors/session.selectors.ts
```diff
@@ -16,7 +16,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses>
  */
 
-import { createFeatureSelector } from '@ngrx/store';
+import { createFeatureSelector, createSelector } from '@ngrx/store';
```
BRUCELLA2, line 19: createSelector is never used

comixed-web/src/app/services/session.service.ts
mcpierce: The goal there is to divide up all of the various parts of the session that will use the messaging bus. Rather than having SessionService have some huge, monolithic method that subscribes to task counts, comic changes, reading list changes, etc., we'll code up each type of subscription as a separate service type (like TaskCountService) that can be started and stopped by SessionService.

That will also make it easier to test them since, like TaskCountService, they are isolated and testable by themselves.

comixed-web/src/app/reducers/messaging.reducer.ts
mcpierce: I don't think so, no. I kept the flag changes simple here, having the actions only flip the one the ones relating to the action itself.

comixed-web/src/app/reducers/messaging.reducer.ts
mcpierce: See the previous answer. Though if there's a logically reason to flip the other flag as well, I'm open to it.

comixed-web/src/app/reducers/messaging.reducer.ts
mcpierce: Ah, good point. When restartMessaging is called it should set startingMessaging to true as well as setting messagingStarted to false. I'll update the code.

comixed-web/src/app/selectors/session.selectors.ts
mcpierce: I've removed it now.

comixed-web/src/app/reducers/messaging.reducer.ts
BRUCELLA2: OK

comixed-web/src/app/reducers/messaging.reducer.ts
BRUCELLA2: OK

comixed-web/src/app/reducers/messaging.reducer.ts
BRUCELLA2: OK

BRUCELLA2 (approved):

Issue #648, opened 2021-03-12
Add security to the messaging layer


mcpierce:
The messaging layer should secure endpoints to prevent misuse.

PR #649, opened 2021-03-12
Added security to the STOMP messaging layer [#648]

mcpierce:
* Attach the web auth token to outbound websocket connections.
 * Generate a principal as a part of this process.

## Status
READY

NO

## Before You Submit Your PR:


comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+  @Autowired private ComiXedUserDetailsService userDetailsService;
+  @Autowired private JwtTokenUtil jwtTokenUtil;
+
+  @Override
+  public Message<?> preSend(final Message<?> message, final MessageChannel channel) {
+    StompHeaderAccessor accessor =
+        MessageHeaderAccessor.getAccessor(message, StompHeaderAccessor.class);
+    if (accessor != null && StompCommand.CONNECT.equals(accessor.getCommand())) {
+      final List<String> authorization = accessor.getNativeHeader(HEADER_STRING);
+      if (authorization != null && !authorization.isEmpty()) {
```
BRUCELLA2: What do you think about using CollectionUtils.isEmpty from the spring framework or CollectionUtils.isNotEmpty from Apache Commons Collections both being Null-safe check ?

Stream and Optional can be used but I'm not sure whether it's clearer with something like : 
Optional<String> optionalToken = Optional.ofNullable(authorization).orElse(Collections.emptyList()).stream().flatMap(Stream::ofNullable).findFirst();
optionalToken.ifPresent(token -> {
        log.trace("Auth token found");
        doCreatePrincipal(accessor, token);
});

comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  void doCreatePrincipal(final StompHeaderAccessor accessor, final String token) {
+    String username = null;
+    try {
+      username = this.jwtTokenUtil.getEmailFromToken(token);
+    } catch (Exception error) {
+      log.error("Unable to extract username from auth token", error);
+    }
+    if (username != null) {
```
BRUCELLA2: Shouldn't we check if the username is blank?

BRUCELLA2 (commented):
Just a question, can you confim why you removed the constant class for authentication and messaging constants ?

comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
mcpierce: I like it! I need to spend more time thinking in streams for things like this. 😄

comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
mcpierce: Yeah, I suppose we should. Normally what should come out of the encoded token is either a value or a null. But to be well-balanced, we can use StringUtils.isEmpty() to be complete.

mcpierce: I've also applied these changes to ComiXedAuthenticationFilter since it's doing similar behavior.

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationFilter.java
```diff
@@ -58,28 +61,29 @@ protected void doFilterInternal(
     String username = null;
     String password = null;
     String authToken = null;
-    if ((header != null) && header.startsWith(TOKEN_PREFIX)) {
+    if (!StringUtils.isEmpty(header) && header.startsWith(TOKEN_PREFIX)) {
```
BRUCELLA2: if we use startsWith(CharSequence str, CharSequence prefix) from Apache Commons (https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.12.0) we can simplify this line (StringUtils from Spring Framework is less complete and there is no startsWith method) : if (StringUtils.startsWith(header, TOKEN_PREFIX)

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationFilter.java
```diff
@@ -58,28 +61,29 @@ protected void doFilterInternal(
...
       authToken = header.replace(TOKEN_PREFIX, "").trim();
       try {
         username = this.jwtTokenUtil.getEmailFromToken(authToken);
       } catch (Exception error) {
         log.error("Unable to extract username from auth token", error);
       }
-    } else if ((header != null) && header.toLowerCase().startsWith("basic")) {
+    } else if ((!StringUtils.isEmpty(header)) && header.startsWith(BASIC_PREFIX)) {
```
BRUCELLA2: Same as before, use Apache commons StringUtils can be better

comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+  static final String TOKEN_PREFIX = "Bearer ";
+
+  @Autowired private ComiXedUserDetailsService userDetailsService;
+  @Autowired private JwtTokenUtil jwtTokenUtil;
+
+  @Override
+  public Message<?> preSend(final Message<?> message, final MessageChannel channel) {
+    StompHeaderAccessor accessor =
+        MessageHeaderAccessor.getAccessor(message, StompHeaderAccessor.class);
+    log.debug("Received STOMP command: {}", StompCommand.valueOf(accessor.getCommand().name()));
```
BRUCELLA2: if accessor is null -> NPE

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationFilter.java
mcpierce: Changed it to match your suggestion.

comixed-rest-api/src/main/java/org/comixedproject/authentication/ComiXedAuthenticationFilter.java
mcpierce: I've changed the code here as well.

comixed-rest-api/src/main/java/org/comixedproject/messaging/ComiXedWebSocketChannelInterceptor.java
mcpierce: I'm not sure that this code will ever get invoked outside of the context of an incoming STOMP command, so accessor should implicitly always be non-null. But, that said, I've added a null check to be safe.

BRUCELLA2 (approved):

Issue #747, opened 2021-05-30
realpath is not available on stock Mac OS.


SinisterStairs:
**Describe The Bug**
macOS cannot run the release's `run.sh` as-is because `realpath` doesn't exist.  (I've tested on 10.14 (Mojave) and 11.4 (Big Sur))

**To Reproduce The Bug**
To fix the problem with Mojave, I installed GNU `readlink`.

Using Homebrew I did `brew install coreutils`.  Personally, I `cp /usr/local/bin/readlink $HOME/bin/` then uninstalled coreutils.

Then I edited  `run.sh` and changed the references of `realpath -s` to `readlink -c`

I'm not sure if this is so much a bug w/ workaround, or some helpful information to document.  Note:  I verified that the release's `run.sh` doesn't work in either Mojave or Big Sur, but my workaround is specifically for Mojave.

Big Sur does have readlink already installed, but it is not GNU's.  Installing GNU readlink in addition to macOS's could have side effects, although by default Homebrew will install to /opt/Homebrew/bin and not override the OS.  But the user's PATH will need to have Homebrew over /usr/bin

mcpierce:
@SinisterStairs Thanks for bringing this to our attention!

I think I've found a reasonable workaround for requiring users to install Homebrew and coreutils, and would appreciate if you could take a look and see if it works for you. If so, I'll add it to the code and push an update.

(edit)

[Here](https://github.com/comixed/comixed/pull/748) is the link to the PR. If you could provide feedback on the solution I'll merge it and put out a new 0.8 release with it.

SinisterStairs:
Well duh!  I should have thought of that myself!  Perfect solution!  (I already had readlink installed for other reasons, so I was looking at it with a blindspot of what was "needed.")

This works on my personal Macbook Pro (Mojave), but not my work Mac (Big Sur).  I'll provide details in the PR.

mcpierce:
@SinisterStairs When you're happy with the resolution, please feel free to close this ticket.

SinisterStairs:
Orirginal problem fixed, thanks!

Note to self: A corner case (for all *nix) exists when the filepath contains special characters, such as spaces.

EDIT: Fixed in pull request #758

PR #748, opened 2021-05-30
Fixed requiring Homebrew coreutils to use scripts on the Mac [#747]

mcpierce:
## Status
READY

NO

## Before You Submit Your PR:
- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/comixed-release/src/main/assembly/scripts/run.sh
+++ b/comixed-release/src/main/assembly/scripts/run.sh
@@ -16,9 +16,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses>
 
-ME=$(realpath -s $0)
+realpath() {
+  [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
+}
+
+ME=$(realpath $0)
 BINDIR=$(dirname ${ME})
-LIBDIR=$(realpath -s ${BINDIR}/../lib)
+LIBDIR=$(realpath ${BINDIR}/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''


SinisterStairs (approved):
This fixes the `realpath` issue for both Mojave (macOS 10.4) and Big Sur (macOS 11.4).  I didn't have a Catalina or pre-Mojave macOS system to test on, but I'd be genuinely surprised if it didn't also work.

EDIT: Emphasis on fixing `realpath` -- not ComiXed.  See general comments.

SinisterStairs:
This patch fixed `run.sh` with Big Sur (macOS 11.4), but `java` itself core dumped.  (No issues with Mojave; Mojave runs ComiXed now.)

I am using OpenJDK 16 on both those machines, as installed by Homebrew; but of note, my Big Sur* MBP also has Apple's M1** chip.  So the JRE core dump could be related to Homebrew not detecting the M1 chip and installing the M1 version of AdoptOpenJDK, or it could be that OpenJDK isn't compatible with the M1 chip.

I'm not going to spend much more time looking into this because I only cared about running ComiXed on my home computer or NAS; and on my work computer, Eclipse is running/debugging Java applications just fine (using its internal JDK).

But if I have run into other general JDK issues on my Big Sur / M1 MBP (which I'd have to resolve for work), I'll update the solution here.

*Mojave was the last macOS to support 32-bit applications (in addition to 64-bit), while Catalina and Big Sur are 64-bit only.  This caused problems for applications which weren't recompiled, but this isn't likely the problem here.

**In late 2020, Apple switched from Intel chips to their own M1 chip.  So in addition to running Big Sur, some MBP may be using the M1 chip rather than Intel.  This has added yet another compatibility vector.

SinisterStairs (changes requested):
Actually this didn't completely fix the functionality of `realpath`:

The `run.sh` script works fine if I invoke it with `./run.sh` but when I tried `bin/run.sh` (from the source directory) I got:

Error: Unable to access jarfile ~/src/comixed-release-0.8.2-1.1/comixed-app-*.jar

So it dropped `bin/` from the path

EDIT:  Running it from my home directory with `./src/comixed-release-0.8.2-1.1/bin/run.sh` the realpath implementation returns my home directory, so it appears the realpath function is just returning the cwd

SinisterStairs (changes requested):
I figured it out:  the realpath function is working, but the bash script needs to call it with just `realpath` instead of `realpath -s`

EDIT:  We ought to do some regression testing to make sure the `realpath` function is compatible with Windows and *nix machines.  Unfortunately both my Windows PC and NAS (*nix) are down atm, but as long as system supports bash, I don't see why it wouldn't.

mcpierce:
> I figured it out: the realpath function is working, but the bash script needs to call it with just `realpath` instead of `realpath -s`
> 
> EDIT: We ought to do some regression testing to make sure the `realpath` function is compatible with Windows and *nix machines. Unfortunately both my Windows PC and NAS (*nix) are down atm, but as long as system supports bash, I don't see why it wouldn't.

I'll post an updated PR with the -s portion removed. I've got Linux boxes I can try them out on as well to be sure the script works there as well, though I agree with you that they should work just fine.

(edit)

Updated PR is pushed. Please take a look and let me know what you think.

mcpierce:
@SinisterStairs Were you able to verify the updated PR as working? If so, please let me know and I'll merge it.

SinisterStairs:
It works for the happy path -- I'll comment in the source, but basically want to escape some variables.

I'm not sure I tested properly:  I downloaded the official release.zip, then cloned and checked out your branch, and replaced the release's run.sh with yours -- that worked in Mojave.

mcpierce:
@SinisterStairs Using the release archive but replacing the launch script should be a perfectly fine test; i.e., the script hasn't changed for the upcoming release.

I'll go ahead and merge this PR then and, if there are further issues, we can re-open the source ticket and take further action. And thank you for getting back to me.

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -16,9 +16,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses>
 
-ME=$(realpath -s $0)
+realpath() {
+  [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
```
SinisterStairs, line 20: EDIT:  The first `$1` needs shell escaping so it works with filepaths containing spaces and other special characters:  `"$1"`

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -16,9 +16,13 @@
...
+}
+
+ME=$(realpath $0)
```
SinisterStairs, line 23: This needs to be shell escaped so it can handle directories with spaces in them (or other special characters):  `realpath "$0"`

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -16,9 +16,13 @@
...
 BINDIR=$(dirname ${ME})
```
SinisterStairs, line 24: Also needs shell escaping, `dirname "${ME}"`

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -16,9 +16,13 @@
...
-LIBDIR=$(realpath -s ${BINDIR}/../lib)
+LIBDIR=$(realpath ${BINDIR}/../lib)
```
SinisterStairs, line 25: Shell escape `"${BINDIR}"`

SinisterStairs (commented):
In addition to those lines, line 30 of the bugfix branch should be escaped: `COMIXED_JAR_FILE="${BINDIR}"/...` (I think...this is where I'm currently stuck with testing.  `JVMOPTIONS=` might need escaping; the `$JAVA ... -jar $COMIXED_JAR_FILE` line is what's failing if there are spaces or special characters in the filepath

SinisterStairs:
I should just make a merge request myself, but anyways, the happy path works

PR #758, opened 2021-06-03
Fixed *nix release script to support directory paths with spaces [#747]

SinisterStairs:
This fixes `run.sh` for ComiXed directory paths that might have spaces in them.  This also fixed a logic bug with testing for the `comixed-*.jar` file.

Please see the my source comments:  There were two weird (`bash`) behaviors I worked around, and they should be tested in Linux.  (My changes were tested in macOS.)

I personally don't have spaces in my ComiXed directory path, and I'm not sure why I thought to test/fix that case.  (And I regret taking the time to do so.  Those weird behaviors and subsequent workarounds took more time than this case deserved!)

### DIFF
--- a/comixed-release/src/main/assembly/scripts/run.sh
+++ b/comixed-release/src/main/assembly/scripts/run.sh
@@ -20,14 +20,14 @@ realpath() {
   [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
 }
 
-ME=$(realpath $0)
-BINDIR=$(dirname ${ME})
-LIBDIR=$(realpath ${BINDIR}/../lib)
+ME=$(realpath "$0")
+BINDIR=$(dirname "${ME}")
+LIBDIR=$(realpath "${BINDIR}"/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''
 JVMOPTIONS=''
-COMIXED_JAR_FILE=${BINDIR}/comixed-app-*.jar
+COMIXED_JAR_FILE=$(echo "${BINDIR}"/comixed-app-*.jar)
 DEBUG=false
 FULL_DEBUG=false
 VERBOSE=false
@@ -74,7 +74,7 @@ if $VERBOSE; then
   set -x
 fi
 
-if [[ -f "${COMIXED_JAR_FILE}" ]]; then
+if [[ ! -f "${COMIXED_JAR_FILE}" ]]; then
   echo "Missing JAR file"
   exit 1
 fi
@@ -112,7 +112,8 @@ fi
 # build a list of JVM arguments
 
 if [[ $LIBDIR ]]; then
-  JVMOPTIONS="${JVMOPTIONS} -classpath ${LIBDIR}"
+  JVMOPTIONS="${JVMOPTIONS} -classpath"
+  CLASSPATH="${LIBDIR}"
 fi
 
-$JAVA ${JVMOPTIONS} -jar ${COMIXED_JAR_FILE} ${JAROPTIONS}
+$JAVA ${JVMOPTIONS} "${CLASSPATH}" -jar "${COMIXED_JAR_FILE}" ${JAROPTIONS}


comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -20,14 +20,14 @@ realpath() {
...
-LIBDIR=$(realpath ${BINDIR}/../lib)
+ME=$(realpath "$0")
+BINDIR=$(dirname "${ME}")
+LIBDIR=$(realpath "${BINDIR}"/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''
 JVMOPTIONS=''
-COMIXED_JAR_FILE=${BINDIR}/comixed-app-*.jar
+COMIXED_JAR_FILE=$(echo "${BINDIR}"/comixed-app-*.jar)
```
SinisterStairs, line 30: This might be specific to macOS's bash: `COMIXED_JAR_FILE="${BINDIR}"/comixed-app-*.jar` would not expand; it would remain a literal `*`, hence my `echo` hack. We should regression test this particular change in Linux.

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -74,7 +74,7 @@ if $VERBOSE; then
   set -x
 fi
 
-if [[ -f "${COMIXED_JAR_FILE}" ]]; then
+if [[ ! -f "${COMIXED_JAR_FILE}" ]]; then
```
SinisterStairs, line 77: This is an existing bug -- I'm not sure how it ever worked. `-f` is for testing "a regular file (not a directory or device file)." We should be testing for when the JAR _doesn't_ exist or when it's _not_ a regular file.

The original line here "worked" prior to my shell-escape fixes (it didn't trigger an error, anyways); the line failed with the shell-escape fixes in this patch.

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -112,7 +112,8 @@ fi
 # build a list of JVM arguments
 
 if [[ $LIBDIR ]]; then
-  JVMOPTIONS="${JVMOPTIONS} -classpath ${LIBDIR}"
+  JVMOPTIONS="${JVMOPTIONS} -classpath"
```
SinisterStairs, line 115: Lines 115-116: Another weird expansion problem that might be specific to macOS's bash (similar to the comment on line 30). Without this hack, either the `JVMOPTIONS` wouldn't have proper parameters (i.e. it'd be one long string) or spaces in the classpath would cause the java invocation to fail. Again, we should regression test this in Linux too.

mcpierce, line 115: I'm not sure I follow the logic here: why not include the LIBDIR value immediately after the -classpath option?

comixed-release/src/main/assembly/scripts/run.sh
mcpierce, line 77: Oooh, thanks for catching that.

mcpierce (changes requested):
Could change the first line to comply with our [commit message convention](https://github.com/comixed/comixed/blob/develop/CONTRIBUTING.md)? Something like:

```
Fixed run.sh for *nix systems [#747]

Enclosed variables in quotes to ensure that spaces would
not affect their usage.
```

And, also, please add yourself to the CONTRIBUTORS.md file.

SinisterStairs (commented):
> Could change the first line to comply with our commit message convention

I'll delete my fork and resubmit it, because I can't figure out how to fix my git commit messages without leaving lots of annoying merges.  I'm mentioning it here because I imagine it will trash this pull request as well.

comixed-release/src/main/assembly/scripts/run.sh
SinisterStairs: This is what happens if I use the original line 115:
```bash
$ ~/src/Co\ Mixed/bin/run.sh -v
+ [[ ! -f /Users/jfine/src/Co Mixed/bin/comixed-app-0.8.2-1.1.jar ]]
+ false
+ false
+ [[ -n '' ]]
+ [[ -n '' ]]
+ [[ -n '' ]]
+ [[ -n '' ]]
+ [[ -n '' ]]
+ [[ -n /Users/jfine/src/Co Mixed/bin/../lib ]]
+ JVMOPTIONS=' -classpath /Users/jfine/src/Co Mixed/bin/../lib'
+ /usr/bin/java -classpath /Users/jfine/src/Co Mixed/bin/../lib -jar '/Users/jfine/src/Co Mixed/bin/comixed-app-0.8.2-1.1.jar'
Error: Could not find or load main class Mixed.bin....lib
```
I've not had this problem with Linux before, so I wonder if it's a weird Mac bash quirk.  (Mojave and Big Sur use bash 3.2.x)

My workaround was to put the LIBDIR (which contains a space in it) all in its variable so I can properly shell escape.  I created a second CLASSPATH variable so if LIBDIR doesn't exist (but was set), no string is substituted.

SinisterStairs: Tested this on a Linux machine with bash 4.1.x -- same problem.

SinisterStairs:
Closing this PR to submit PR [#759]  which doesn't have all the weird commit logs / better follows contributing conventions (I hope).

PR #759, opened 2021-06-04
Fixed run.sh for *nix releases [#747]

SinisterStairs:
* Enclosed variables in quotes to permit directory paths with spaces
 * Fixed a logic flaw with how the JAR file was being tested
 * Added self to contributors

### DIFF
--- a/comixed-release/src/main/assembly/scripts/run.sh
+++ b/comixed-release/src/main/assembly/scripts/run.sh
@@ -20,14 +20,14 @@ realpath() {
   [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
 }
 
-ME=$(realpath $0)
-BINDIR=$(dirname ${ME})
-LIBDIR=$(realpath ${BINDIR}/../lib)
+ME=$(realpath "$0")
+BINDIR=$(dirname "${ME}")
+LIBDIR=$(realpath "${BINDIR}"/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''
 JVMOPTIONS=''
-COMIXED_JAR_FILE=${BINDIR}/comixed-app-*.jar
+COMIXED_JAR_FILE=$(echo "${BINDIR}"/comixed-app-*.jar)
 DEBUG=false
 FULL_DEBUG=false
 VERBOSE=false
@@ -74,7 +74,7 @@ if $VERBOSE; then
   set -x
 fi
 
-if [[ -f "${COMIXED_JAR_FILE}" ]]; then
+if [[ ! -f "${COMIXED_JAR_FILE}" ]]; then
   echo "Missing JAR file"
   exit 1
 fi
@@ -112,7 +112,8 @@ fi
 # build a list of JVM arguments
 
 if [[ $LIBDIR ]]; then
-  JVMOPTIONS="${JVMOPTIONS} -classpath ${LIBDIR}"
+  JVMOPTIONS="$JVMOPTIONS -classpath"
+  CLASSPATH="${LIBDIR}"
 fi
 
-$JAVA ${JVMOPTIONS} -jar ${COMIXED_JAR_FILE} ${JAROPTIONS}
+$JAVA ${JVMOPTIONS} "${CLASSPATH}" -jar "${COMIXED_JAR_FILE}" ${JAROPTIONS}
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -10,3 +10,4 @@ what it is today.
  * Pidockmedia <pidockmedia@gmail.com>
  * Frédéric Fouret (BRUCELLA2) <fouretfrederic@gmail.com>
  * Arturo Pérez <theotocopulitos@gmail.com>
+ * J Fine <jfine@jfine.com>


comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -20,14 +20,14 @@ realpath() {
...
-LIBDIR=$(realpath ${BINDIR}/../lib)
+ME=$(realpath "$0")
+BINDIR=$(dirname "${ME}")
+LIBDIR=$(realpath "${BINDIR}"/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''
 JVMOPTIONS=''
-COMIXED_JAR_FILE=${BINDIR}/comixed-app-*.jar
+COMIXED_JAR_FILE=$(echo "${BINDIR}"/comixed-app-*.jar)
```
SinisterStairs, line 30: `BINDIR` needs to be enclosed in quotes to account for directory paths with spaces.  However, enclosing it in quotes means the `*` doesn't get expanded and becomes a literal.  Therefore I use the echo to escape `BINDIR`, then append the `*` for expansion.

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -20,14 +20,14 @@ realpath() {
...
-BINDIR=$(dirname ${ME})
-LIBDIR=$(realpath ${BINDIR}/../lib)
+ME=$(realpath "$0")
+BINDIR=$(dirname "${ME}")
+LIBDIR=$(realpath "${BINDIR}"/../lib)
 
 JAVA=$(which java)
 JAROPTIONS=''
 JVMOPTIONS=''
-COMIXED_JAR_FILE=${BINDIR}/comixed-app-*.jar
```
SinisterStairs, line 30: `BINDIR` needs to be enclosed in quotes to allow directory paths with spaces. However, enclosing it in quotes like `"${BINDIR}"/comixed-app-*.jar` prevents(!) file expansion and the `*` becomes a literal character...

comixed-release/src/main/assembly/scripts/run.sh
SinisterStairs, line 30: ...Therefore I use the `echo` to escape `BINDIR`, then append the `*` for expansion.

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -74,7 +74,7 @@ if $VERBOSE; then
   set -x
 fi
 
-if [[ -f "${COMIXED_JAR_FILE}" ]]; then
+if [[ ! -f "${COMIXED_JAR_FILE}" ]]; then
```
SinisterStairs, line 77: I'm not sure how this ever worked:  `-f` tests "if file exists and is a regular file" (not a directory or device).  So the previous code was testing, "If JAR exists, exit because the JAR is missing."  I'm guessing it worked because the `COMIXED_JAR_FILE`'s `*` was being treated like a literal, so by failing, it succeeded?

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -112,7 +112,8 @@ fi
 # build a list of JVM arguments
 
 if [[ $LIBDIR ]]; then
-  JVMOPTIONS="${JVMOPTIONS} -classpath ${LIBDIR}"
+  JVMOPTIONS="$JVMOPTIONS -classpath"
+  CLASSPATH="${LIBDIR}"
 fi
 
-$JAVA ${JVMOPTIONS} -jar ${COMIXED_JAR_FILE} ${JAROPTIONS}
```
SinisterStairs, line 118: This line fails when `LIBDIR` has a space in it (e.g. `/home/John Doe/comixed`) because the shell interprets the it as `-classpath` `/home/John` `Doe/comixed` (3 parameters).

See lines 116 and 119 in the patch for how I worked around this...

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -112,7 +112,8 @@ fi
 # build a list of JVM arguments
 
 if [[ $LIBDIR ]]; then
-  JVMOPTIONS="${JVMOPTIONS} -classpath ${LIBDIR}"
+  JVMOPTIONS="$JVMOPTIONS -classpath"
+  CLASSPATH="${LIBDIR}"
```
SinisterStairs, line 116: See my comment in [line 118 of the original source](https://github.com/comixed/comixed/pull/759#discussion_r645262486) for why this line was a problem.

If `LIBDIR` exists, I make a copy of its value in `CLASSPATH`.  Note that if `LIBDIR` _doesn't_ exist:  `JVMOPTIONS` will be unchanged (i.e. won't contain `-classpath`), and the `CLASSPATH` will be an empty string in line 119 below.  (i.e. `-classpath $LIBDIR` won't be included).

comixed-release/src/main/assembly/scripts/run.sh
SinisterStairs, line 118: This line fails when `LIBDIR` has a space in it (e.g. `/home/John Doe/comixed`) because the shell interprets the it as `-classpath`, `/home/John`, `Doe/comixed` (3 parameters).

I can't enclose `JVMOPTIONS` in quotes because then the shell interprets it as `-classpath /home/John Doe/comixed` (1 parameter, when in fact it should be 2).

See lines 116 and 119 in the patch for how I worked around this...

comixed-release/src/main/assembly/scripts/run.sh
```diff
@@ -112,7 +112,8 @@ fi
...
 fi
 
-$JAVA ${JVMOPTIONS} -jar ${COMIXED_JAR_FILE} ${JAROPTIONS}
+$JAVA ${JVMOPTIONS} "${CLASSPATH}" -jar "${COMIXED_JAR_FILE}" ${JAROPTIONS}
```
SinisterStairs, line 119: As noted above, if `LIBDIR` didn't exist, `JVMOPTIONS` is unchanged and `CLASSPATH` is an empty string.

If `LIBDIR` does exist, `JVMOPTIONS` suffixes the `-classpath` and here I can enclose `CLASSPATH` (`LIBDIR`) in quotes to properly handle directory paths with spaces in them.

SinisterStairs:
`run.sh` failed on both macOS and Linux when the directory path contained spaces.  (e.g. `/home/John Doe/comixed`).

* I tested this fix in macOS (Mojave and Big Sur), which both use bash 3.2.x
* I also tested this fix on a Linux box with bash 4.1.x

comixed-release/src/main/assembly/scripts/run.sh
mcpierce: That sounds right.

mcpierce (approved):

Issue #200, opened 2020-03-24
Display some details on a comic in the popup menu


mcpierce:
When right clicking on a comic, show a few pieces of high-level information on the issue:
 * The publisher, series, volume and issue number.
 * The cover date.
 * The added date.
 * The last read date (if applicable).
 * A link to download the comic

Then attach the context menu to this dialog.

PR #766, opened 2021-06-13
Added some dropdown details for comic cards [#200]

mcpierce:
## Status
READY/IN DEVELOPMENT/HOLD

NO

## Before You Submit Your PR:

### DIFF
--- a/comixed-web/src/app/comic-book/comic-book.module.ts
+++ b/comixed-web/src/app/comic-book/comic-book.module.ts
@@ -83,6 +83,7 @@ import {
 import { UpdateComicInfoEffects } from '@app/comic-book/effects/update-comic-info.effects';
 import { MatSortModule } from '@angular/material/sort';
 import { CoreModule } from '@app/core/core.module';
+import { MatGridListModule } from '@angular/material/grid-list';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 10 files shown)


BRUCELLA2 (commented):
When we hover the expansion panel header we have white text on white background.
Maybe it would be interesting to add a piece of css for the hover with a secondary color (to define) : 
.cx-primary-dark-background:hover {
  background: var(--secondary-darker-color);
  color: var(--text-secondary-darker-color);
}

mcpierce:
> When we hover the expansion panel header we have white text on white background.
> Maybe it would be interesting to add a piece of css for the hover with a secondary color (to define) :
> .cx-primary-dark-background:hover {
> background: var(--secondary-darker-color);
> color: var(--text-secondary-darker-color);
> }

Doing this affects everything that uses cx-primary-dark-background when on that view, including the toolbar.

Instead I tried to do a more targeted use by styling .mat-expansion-panel:hover and .mat-expansionpanel-content and things look good during hover and dropdown. But then closing the expansion panel turns it to white-on-white and I can't seem to find the styling to override it.

(edit)

Found that adding !important to the global styling fixed the issue with the expansion panel going to white-on-white. So now it's consistent between the expanded and closed states and looks good. Thanks for catching that!

mcpierce:
@BRUCELLA2 Is this PR acceptable?

PR #842, opened 2021-08-18
Fix issues with SonarCloud

BRUCELLA2:
Fix issues with SonarCloud

## Status
WIP

Unit tests on java code do not seem to be taken into account. This PR try to fix this. WIP

### DIFF
--- a/comixed-release/pom.xml
+++ b/comixed-release/pom.xml
@@ -67,19 +67,6 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.jacoco</groupId>
-        <artifactId>jacoco-maven-plugin</artifactId>
-        <version>0.8.5</version>
-        <executions>
-          <execution>
-            <phase>verify</phase>
-            <goals>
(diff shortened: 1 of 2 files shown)


comixed-services/src/main/java/org/comixedproject/service/comic/ComicFormatService.java
```diff
@@ -38,6 +38,10 @@
 
   public List<ComicFormat> getAll() {
     log.debug("Getting all comic formts");
-    return this.comicFormatRepository.findAll();
+    if (true) {
```
mcpierce: I don't see a reason, even to make SonarCloud happy, to add this sort of code.

mcpierce (changes requested):

mcpierce (changes requested):
Please also make sure to 1) squash all commits together into a single commit for the PR and 2) make sure the commit message follows the guide for commit messages.

mcpierce (approved):

Issue #846, opened 2021-08-19
Administrators can mark one or more comics for deletion


mcpierce:
As an administrator, I need to be able to mark comics that I no longer want in my library.

# Preconditions
1. The user is an administrator.
# User Story
1. The user right-clicks on a comic.
1. The user selects the "Mark for deletion" menu item.
1. The user confirms the marking.
# Postconditions
1. The comic is marked for deletion in the database.
# Alternate User Stories
1. The user selects "Unmark for deletion".
   1. The item is unmarked for deletion.
1. The user selects multiple items before displaying the popup menu.
1. The user clicks the mark for deletion toolbar button.

PR #851, opened 2021-08-20
Added menu items to delete/undelete comics [#846]

mcpierce:
* Added a new feature to the frontend to set the deleted state.
 * Modified the REST API to not return a response value.
 * Added actions to handle the state changes.
 * Renamed the tasks and encoders for performing the changes.
 * Removed deleting the physical file from the marking tasks.

## Status
READY

NO

## Before You Submit Your PR:


comixed-state/src/test/java/org/comixedproject/state/comic/actions/MarkComicForRemovalActionTest.java
```diff
@@ -0,0 +1,39 @@
...
+        .thenReturn(comic);
+  }
+
+  @Test
+  public void testExecute() {
+    Mockito.doNothing().when(comic).setDateDeleted(deletedDateCaptor.capture());
+
+    action.execute(context);
+
+    Mockito.verify(comic, Mockito.times(1)).setDateDeleted(deletedDateCaptor.getValue());
```
BRUCELLA2, line 41: for more security, I would test if deletedDateCaptor.getValue() is not null

BRUCELLA2 (commented):
Just a small comment on a test otherwise it's ok for me.

comixed-state/src/test/java/org/comixedproject/state/comic/actions/MarkComicForRemovalActionTest.java
mcpierce: Agreed. I've added an assertion to make sure it's not null.

BRUCELLA2 (approved):

Issue #905, opened 2021-09-15
Provide translations for the 0.10 release stream


mcpierce:
Provide translations for the project into the following supported languages:

 * Engish - Default
 * French - @BRUCELLA2 
 * Spanish - HELP NEEDED
 * Portuguese - HELP NEEDED
 * German - HELP NEEDED

PR #928, opened 2021-09-20
Added French translation version 0.10 [#905]

BRUCELLA2:
## Status
READY

NO

## Before You Submit Your PR:
- [ ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/comixed-webui/src/assets/i18n/en/admin.json
+++ b/comixed-webui/src/assets/i18n/en/admin.json
@@ -21,7 +21,7 @@
       "renaming-rule-format": "The comic format",
       "renaming-rule-issue-number": "The issue number",
       "renaming-rule-published-month": "The month published",
-      "renaming-rule-published-year": "The name of the publisher",
+      "renaming-rule-published-year": "The year published",
       "renaming-rule-publisher": "The name of the publisher",
       "renaming-rule-series": "The name of the series",
       "renaming-rule-volume": "The volume",
--- a/comixed-webui/src/assets/i18n/fr/app.json
+++ b/comixed-webui/src/assets/i18n/fr/app.json
@@ -55,14 +55,14 @@
   "home": {
     "label": {
     },
-    "title": "Home"
+    "title": "Accueil"
   },
   "menu": {
     "language-menu": {
       "root": "Choix de la langue"
     },
     "logging-menu": {
-      "root": "Logging",
+      "root": "Journalisation",
       "1": "Tout (Verbeux)",
       "2": "Trace (Encore plus)",
       "3": "Debug (Beaucoup)",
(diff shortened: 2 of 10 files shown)


comixed-webui/src/assets/i18n/en/admin.json
```diff
@@ -21,7 +21,7 @@
       "renaming-rule-format": "The comic format",
       "renaming-rule-issue-number": "The issue number",
       "renaming-rule-published-month": "The month published",
-      "renaming-rule-published-year": "The name of the publisher",
+      "renaming-rule-published-year": "The year published",
```
mcpierce, line 24: Thanks for catching that! To keep the other translations consistent, would you mind adding this same change to the other admin.json files? That way other translators will see what was intended.

comixed-webui/src/assets/i18n/fr/comic-book.json
```diff
@@ -10,12 +10,12 @@
   },
```
mcpierce, line 10: The archive types entries for "all" and "select" need to be translated.

comixed-webui/src/assets/i18n/fr/app.json
```diff
@@ -111,7 +111,7 @@
       "import-comics": "Bandes Dessinées importées",
       "location-collection": "Lieux",
       "publisher-collection": "Éditeurs",
-      "reading-lists": "Reading Lists",
```
mcpierce, line 114: Line 58, "title", should be translated.
Line 65, "root", should be translated.

mcpierce (changes requested):
A couple of changes requested and this is ready to go. Thanks for the prompt PR!

comixed-webui/src/assets/i18n/fr/app.json
BRUCELLA2: The terms Home and Logging were difficult to translate. Home can still be translated as "Accueil", I hadn't thought of that. On the other hand Logging, the translation would be "Journalisation" but I think I have never heard it.We always talk about "log", "logging system". But Quebecers should find it more natural ;)

mcpierce: Would it make more sense to change the base to "Logs" so it's a noun rather than a verb? Then the translation could be "Journals" and be more accurate, and the entries would be more consistent since they would all be clear references to an object.

mcpierce (changes requested):

comixed-webui/src/assets/i18n/fr/app.json
BRUCELLA2: The term I often use and which seems to me the most adapted is logging levels (levels of logs in French).  But it is also often used logging as you can see here: https://www.jmdoudoux.fr/java/dej/chap-logging.htm
So in my opinion logging sounds good or logging levels.

mcpierce (approved):

PR #1194, opened 2022-03-27
Added using english as the reference language [#1193]

### LINKED ISSUE #1193: English is the reference translation fileset.
As a translation contributor, I want to provide a complete translation for each release. To do this, I would like to take the english translation, copy it into the appropriate directory, and translate the entries to my target language.

mcpierce:
Closes #1193 

## Status
READY

NO

## Before You Submit Your PR:


BRUCELLA2:
If I understand correctly the purpose of this PR is to add a new language to Comixed. It's a good idea to have an option to not have all the languages available and only the relevant ones.
On the other hand, I'm not sure that leaving the translation entirely up to the user (administrator) is a good idea. It would be less work for the team but I'm not sure that many users will have the courage to translate everything.
Maybe we should make a repository with a directory for each version and a subdirectory for each language with the translation files that could be retrieved and updated by the community for example.

mcpierce:
> If I understand correctly the purpose of this PR is to add a new language to Comixed. It's a good idea to have an option to not have all the languages available and only the relevant ones. On the other hand, I'm not sure that leaving the translation entirely up to the user (administrator) is a good idea. It would be less work for the team but I'm not sure that many users will have the courage to translate everything. Maybe we should make a repository with a directory for each version and a subdirectory for each language with the translation files that could be retrieved and updated by the community for example.

I'm not sure I follow what you're saying. We would move the translations to an external project? How would we integrate the translation back into the main code?

BRUCELLA2:
Um, I think I went a bit far in my head. Forget about creating a repository ....
But I still think it's not a good thing to delete the different translations even if it means extra work. I'm committed to doing the French at least ;)

mcpierce:
> Um, I think I went a bit far in my head. Forget about creating a repository .... But I still think it's not a good thing to delete the different translations even if it means extra work. I'm committed to doing the French at least ;)

Understood. I'll decline this PR then and close the ticket.

Issue #1210, opened 2022-03-31
CTRL-C does not shut down the application.


mcpierce:
**Describe The Bug**
Hitting CTRL-C does not kill the app the first time. It has to be pressed multiple times, or else the Java process identified and killed, to shut the app down.

**To Reproduce The Bug**
Steps to reproduce the behavior. Please be as complete as possible and describe the exact steps and values used to cause the error:
1. Start the application.
1. Press CTRL-C.
1. The process does not fully exit.

**Expected Behavior**
1. The process should shutdown and exit.

PR #1222, opened 2022-04-07
Fixed Shutdown issue [#1210]

jmcrfp:
## Status
READY

NO

## Before You Submit Your PR:
- [ ] New feature (non-breaking change which adds functionality)
- [ ] All new and existing tests passed.

### DIFF
--- a/comixed-app/src/main/java/org/comixedproject/ComiXedApp.java
+++ b/comixed-app/src/main/java/org/comixedproject/ComiXedApp.java
@@ -52,7 +52,6 @@ public class ComiXedApp implements CommandLineRunner, ApplicationContextAware {
     log.info("Closing application context");
     ((ConfigurableApplicationContext) this.applicationContext).close();
     log.info("Shutting down");
-    System.exit(0);
   }
 
   @Override


comixed-app/src/main/java/org/comixedproject/ComiXedApp.java
```diff
@@ -52,7 +52,6 @@ public void onDestroy() {
     log.info("Closing application context");
     ((ConfigurableApplicationContext) this.applicationContext).close();
     log.info("Shutting down");
-    System.exit(0);
```
mcpierce, line 55: Does simply closing the application context shutdown the application, or do we need to take other steps as well?

jmcrfp: Yes, it should. That's the documentation said.
I also looked into the threads and the they were looked because the system.exit was waiting for Spring exit.

mcpierce: Okay.

comixed-app/src/main/java/org/comixedproject/ComiXedApp.java
```diff
@@ -52,7 +52,6 @@ public void onDestroy() {
     log.info("Closing application context");
     ((ConfigurableApplicationContext) this.applicationContext).close();
     log.info("Shutting down");
```
mcpierce, line 54: Since we're removing the following line, let's also delete this one log.info() call.

mcpierce (changes requested):
One change and then we can merge it.

comixed-app/src/main/java/org/comixedproject/ComiXedApp.java
jmcrfp: We should leave the line.
You will see now that after that line more will appear in the log(from Spring), so it should be there.

mcpierce: Fair enough.

mcpierce (approved):

PR #1229, opened 2022-04-13
Added a comic state widget to the home page [#1224]

### LINKED ISSUE #1224: The home page shows the comic state distribution
As an admin, I'd like to see a distribution of comics in my library across all of the defined states.

mcpierce:
Moved the library statistics display to its own component,
and added a new component for the comic state distribution.

Changed the home page layout to use a tile system.

Also fixed a bug that kept teams from being shown.

Closes #1224 

## Status
READY

NO

## Before You Submit Your PR:

### DIFF
--- a/comixed-webui/src/assets/i18n/pt/app.json
+++ b/comixed-webui/src/assets/i18n/pt/app.json
@@ -56,21 +56,26 @@
   "home": {
     "label": {
       "character": "Personagem",
-      "comics-by-character": "Comics Por Personagem",
-      "comics-by-location": "Comics Por Localização",
-      "comics-by-publisher": "Comics Por Editora",
-      "comics-by-series": "Comics Por Série",
-      "comics-by-story": "Comics Por Estória",
-      "comics-by-team": "Comics Por Equipa",
-      "comic-count": "# de Comics",
+      "comic-state-added": "Added",
+      "comic-state-changed": "Changed",
+      "comic-state-deleted": "Deleted",
+      "comic-state-stable": "Stable",
+      "comic-state-unprocessed": "Unprocessed",
       "location": "Localização",
       "publisher": "Editora",
+      "select-collection-type": "Collection Type",
       "series": "Série",
       "story": "Estória",
       "team": "Equipa"
     },
+    "legend": {
+      "collections-top-entries-x-axis": "Number Of Comics",
+      "comic-state-x-axis": "Number Of Comics In Each State"
+    },
     "tab-title": "Início",
     "title": {
+      "comic-collections-chart": "Top Collection Entries",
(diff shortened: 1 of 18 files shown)


comixed-webui/src/assets/i18n/pt/app.json
```diff
@@ -56,21 +56,26 @@
...
+    "legend": {
+      "collections-top-entries-x-axis": "Number Of Comics",
+      "comic-state-x-axis": "Number Of Comics In Each State"
+    },
     "tab-title": "Início",
     "title": {
+      "comic-collections-chart": "Top Collection Entries",
+      "comic-state-chart": "Comics By State",
       "comic-statistics": "Estatísticas da Biblioteca de Comics",
       "loading-comics": "Carregando Biblioteca..."
```
jmcrfp, line 80: If you want you can add the portuguese translation also

"comic-state-added": "Adicionados",
"comic-state-changed": "Alterados",
"comic-state-deleted": "Apagados",
"comic-state-stable": "Estáveis",
"comic-state-unprocessed": "Por Processar",
"select-collection-type": "Tipo De Coleção",

"collections-top-entries-x-axis": "Quantidade De Comics",
"comic-state-x-axis": "Quantidade De Comics Por Estado"

"comic-collections-chart": "Principais Entradas Da Coleção",
"comic-state-chart": "Comics Por Estado",

jmcrfp:
Everything looks nice, but when I was looking I found something that is not so great but it was probably already there:
So the right box (Comics by state) doesn't show the right border. 
And that's probably because the browser vertical scroll bar appears.
And the reason the it appears is because the default menu side bar vertical size is bigger than the page it self. (I know the menu is bigger that normal page size and it has it own vertical scroll bar, and that's fine) But now I need to scroll 2 bars to see the bottom of the menu.

comixed-webui/src/assets/i18n/pt/app.json
mcpierce: Thanks. Though I'm going to leave translations until the 1.1 release. That way we keep the translations all together.

mcpierce:
> Everything looks nice, but when I was looking I found something that is not so great but it was probably already there: So the right box (Comics by state) doesn't show the right border. And that's probably because the browser vertical scroll bar appears. And the reason the it appears is because the default menu side bar vertical size is bigger than the page it self. (I know the menu is bigger that normal page size and it has it own vertical scroll bar, and that's fine) But now I need to scroll 2 bars to see the bottom of the menu.

I think you're right about it likely already being there. We should probably do some tweaks in a separate commit to clean up the layout. Right now the main application container is set to 100% width. I'll start up a thread in the dev list for how to properly address that.

PR #1263, opened 2022-05-02
Added linking the collections chart to the collection page [#1256]

### LINKED ISSUE #1256: Clicking on an collection entry type on the home page, the user is taken to that collections view.
As a user when viewing a collection (such as publishers, series, etc) and I click on an entry, the app should take me to the details view for that specific collection.

mcpierce:
Closes #1256 

## Status
READY

NO

## Before You Submit Your PR:

### DIFF
--- a/comixed-webui/src/app/collections/pages/collection-detail/collection-detail.component.ts
+++ b/comixed-webui/src/app/collections/pages/collection-detail/collection-detail.component.ts
@@ -109,9 +109,11 @@ export class CollectionDetailComponent implements OnInit, OnDestroy {
             this.comics = entries.filter(comic => {
               switch (this.collectionType) {
                 case CollectionType.PUBLISHERS:
-                  return comic.publisher === this.collectionName;
+                  return (
+                    (comic.publisher || '[UNKNOWN]') === this.collectionName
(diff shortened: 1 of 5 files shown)


BRUCELLA2 (changes requested):
The case 'UNKNOWN' is not properly managed.
We have an empty page when we click on unknown publisher (or other collection type). We need to had this option in the filter in the CollectionDetailComponent.
Something like this : 
```
this.comics = entries.filter(comic => {
              switch (this.collectionType) {
                case CollectionType.PUBLISHERS:
                  if (this.collectionName === 'UNKNOWN') {
                    return comic.publisher === null;
                  } else {
                    return comic.publisher === this.collectionName;
                  }
```

BRUCELLA2:
When you click on unknown nothing happens now. Is this what is desired? Wouldn't it be interesting to find out which elements of the collection are unknown to complete the information?

mcpierce:
I believe the only two collections that'll need to support null is publishers and series. I'm trying out replacing null with '[NULL]' and then using that when filtering to look for issues with null.

BRUCELLA2 (approved):

PR #1396, opened 2022-08-16
Added maintaining comic book selections on the server [#1283]

### LINKED ISSUE #1283: The list of selected comics is stored in a server-side session.
As a user, I would like the list of comics I've selected to be stored on the server during my session. If I reload the browser, the selections are restored to the browser.

mcpierce:
* Added new REST APIs to update and clear the selections.
 * Changed the selection feature to call the REST APIs.
 * Removed the old library selection feature.
 * Enabled JDBC sessions.
 * Added the selected ids to the remote library state.

Closes #1283 

## Status
READY

NO

## Before You Submit Your PR:

### DIFF
--- a/comixed-rest/src/test/java/org/comixedproject/rest/library/LibrarySelectionsControllerTest.java
+++ b/comixed-rest/src/test/java/org/comixedproject/rest/library/LibrarySelectionsControllerTest.java
@@ -1,0 +1,110 @@
+/*
+ * ComiXed - A digital comic book library management application.
+ * Copyright (C) 2022, The ComiXed Project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses>
+ */
+
+package org.comixedproject.rest.library;
+
+import static junit.framework.TestCase.assertNotNull;
+import static junit.framework.TestCase.assertSame;
+import static org.comixedproject.rest.library.LibrarySelectionsController.LIBRARY_SELECTIONS;
+
+import java.util.Set;
+import javax.servlet.http.HttpSession;
+import org.comixedproject.model.net.library.SelectComicBooksRequest;
+import org.comixedproject.service.library.LibrarySelectionException;
+import org.comixedproject.service.library.LibrarySelectionService;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LibrarySelectionsControllerTest {
+  @InjectMocks private LibrarySelectionsController controller;
+  @Mock private HttpSession httpSession;
+  @Mock private LibrarySelectionService librarySelectionService;
+  @Mock private Set<Long> selectedComicIds;
+  @Mock private Set<Long> incomingComicBookIds;
+  @Mock private Set<Long> existingSelections;
+
+  @Test(expected = LibrarySelectionException.class)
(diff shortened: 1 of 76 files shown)


comixed-rest/src/test/java/org/comixedproject/rest/library/LibrarySelectionsControllerTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LibrarySelectionsControllerTest {
+  private static final String TEST_EMAIL = "reader@comixedproject.org";
```
BRUCELLA2: never used

comixed-services/src/main/java/org/comixedproject/service/library/LibrarySelectionService.java
```diff
@@ -0,0 +1,89 @@
...
+    Set<Long> result = existing;
+    if (result == null) {
+      log.trace("Creating new selection set");
+      result = new HashSet<>();
+    }
+    if (adding) {
+      log.debug(
+          "Adding {} comic{} from selection set", incoming.size(), incoming.size() == 1 ? "" : "s");
+      for (final Long idIterator : incoming) {
+        final Long id = idIterator;
```
BRUCELLA2: we can use idIterator directly

comixed-services/src/main/java/org/comixedproject/service/library/LibrarySelectionService.java
```diff
@@ -0,0 +1,89 @@
...
+      }
+      return result;
+    } else {
+      if (existing == null || existing.isEmpty()) {
+        log.debug("No existing selections to remove");
+        return existing;
+      }
+
+      for (final Long idIterator : incoming) {
+        final Long id = idIterator;
```
BRUCELLA2: we can use idIterator directly

comixed-services/src/test/java/org/comixedproject/service/library/LibrarySelectionServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+public class LibrarySelectionServiceTest {
+  private static final Long TEST_INCOMING_COMIC_ID = 717L;
+  private static final Long TEST_EXISTING_COMIC_ID = 129L;
+
+  @InjectMocks private LibrarySelectionService service;
+  @Mock private ComicBookService comicBookService;
+  @Mock private ComicBook comicBook;
+
+  private Set<Long> incomingIds = new HashSet<>();
+  private Set<Long> existingIds = new HashSet<>();
```
BRUCELLA2, line 46: can be final.

BRUCELLA2 (changes requested):

comixed-rest/src/test/java/org/comixedproject/rest/library/LibrarySelectionsControllerTest.java
mcpierce: Removed it.

comixed-services/src/main/java/org/comixedproject/service/library/LibrarySelectionService.java
mcpierce: Resolved by renaming idIterator => id and removed the local variable.

comixed-services/src/main/java/org/comixedproject/service/library/LibrarySelectionService.java
mcpierce: Same as the previous fix.

comixed-services/src/test/java/org/comixedproject/service/library/LibrarySelectionServiceTest.java
mcpierce: Should instance variables be marked final? We don't do that elsewhere in the codebase.

BRUCELLA2: That's right. I've gotten into the habit now of setting all variables to final by default. So if I reassign my variable, the IDE alerts me and I either revise my code or remove the final if necessary. So it's an extra security and it allows to see also from the declaration that the variable will not be reassigned even locally. It also seems to me that it is a performance optimization too. But on this last point, I have just learned that it is not necessary to use final because there is since Java 8 the Effectively Final which allows the compiler to consider the variable as final if it is not reassigned.

I am not expert enough to decide what is better, so I validate the code as it is :) . To think in the future if it is better or not to explicitly declare final all the variables if they are not reassigned.

BRUCELLA2 (approved):

PR #2769, opened 2026-08-30
Changed JWT dependencies to v0.13.0 [#2424]

### LINKED ISSUE #2424: Upgrade JSON Web Tokens to the latest version
Currently we're using v0.9.1, and the latest version as of today is v0.13.0. We should upgrade to the newer release.

zoomgy:
Upgraded the io.jsonwebtoken library from version 0.9.1 to 0.13.0 inside the comixed-auth module to fulfill security requirements. Closes #2424.

- Replaced the legacy unified jar with jjwt-api, jjwt-impl, and jjwt-jackson.
- Refactored JwtTokenUtil to use modern verifyWith and claim chains.
- Updated JwtTokenUtilTest with a secure 256-bit signing mock key.
- Updated the default JWT signing key to meet the minimum 32-byte requirement.
- Moved the JWT signing key configuration to application.properties.
- Added contributor information.

## Status
READY

NO

## Before You Submit Your PR:
- [x] Updates runtime dependencies but does not add new functionality
- [x] My code follows the code style of this project.
- [x] My read of the **CONTRIBUTING** document is complete.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/comixed-app/src/main/resources/comixed-core.properties
+++ b/comixed-app/src/main/resources/comixed-core.properties
@@ -8,7 +8,6 @@ server.port=7171
 server.servlet.context-path=/
 spring.h2.console.enabled=false
 spring.h2.console.path=/dbconsole
-comixed.auth.jwt-signing-key=comixed-project
 
 # ssl configuration
 server.ssl.key-store-type=PKCS12
--- a/comixed-auth/src/test/java/org/comixedproject/auth/JwtTokenUtilTest.java
+++ b/comixed-auth/src/test/java/org/comixedproject/auth/JwtTokenUtilTest.java
@@ -33,7 +33,7 @@ class JwtTokenUtilTest {
     Mockito.when(user.getEmail()).thenReturn(TEST_EMAIL);
     Mockito.when(userDetails.getUsername()).thenReturn(TEST_EMAIL);
 
-    tokenUtil.signingKey = "the-testing-key";
+    tokenUtil.signingKey = "comixed-secured-jwt-signing-key!";
     this.token = tokenUtil.doGenerateToken(TEST_EMAIL);
   }
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,3 +15,4 @@ what it is today.
  * Heather Anderson <heath@odysseus.anderson.name>
  * Sathwik Hejamady Bhat <sathwikhbhat@gmail.com>
  * Joran Van Belle <joran.vanbelle@live.be>
+ * Ayush Thakur <ayushwork0@gmail.com>
--- a/comixed-app/src/main/resources/application.properties
+++ b/comixed-app/src/main/resources/application.properties
@@ -42,3 +42,4 @@ comixed.batch.thread-pool-size=3
 logging.file.name=${user.home}/.comixed/comixed.log
 logging.level.root=OFF
 logging.level.org.comixedproject=INFO
+comixed.auth.jwt-signing-key=comixed-secured-jwt-signing-key!
--- a/comixed-auth/pom.xml
+++ b/comixed-auth/pom.xml
@@ -20,8 +20,20 @@
     </dependency>
     <dependency>
       <groupId>io.jsonwebtoken</groupId>
-      <artifactId>jjwt</artifactId>
-      <version>0.9.1</version>
+      <artifactId>jjwt-api</artifactId>
+      <version>0.13.0</version>
+    </dependency>
+    <dependency>
+      <groupId>io.jsonwebtoken</groupId>
+      <artifactId>jjwt-impl</artifactId>
+      <version>0.13.0</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.jsonwebtoken</groupId>
+      <artifactId>jjwt-jackson</artifactId>
+      <version>0.13.0</version>
+      <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>org.comixedproject</groupId>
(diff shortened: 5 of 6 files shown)


comixed-auth/src/test/java/org/comixedproject/auth/JwtTokenUtilTest.java
```diff
@@ -33,7 +33,7 @@ void setUp() {
     Mockito.when(user.getEmail()).thenReturn(TEST_EMAIL);
     Mockito.when(userDetails.getUsername()).thenReturn(TEST_EMAIL);
 
-    tokenUtil.signingKey = "the-testing-key";
+    tokenUtil.signingKey = "comixed-project-secure-test-key-32";
```
mcpierce: @zoomgy Would you recommend making this a property rather than a hard-coded string value? Are there any concerns if we did that?

mcpierce (changes requested):
@zoomgy All looks good, thank you! Two points:
1. please answer the one question on the PR, and
1. please add yourself to the CONTRIBUTORS.md file.

After these are done, I'll merge this PR.

comixed-auth/src/test/java/org/comixedproject/auth/JwtTokenUtilTest.java
zoomgy: Answer for the two points:

1. Yes, externalizing it could make it easier to change. However, because this is a test-only key and isn't used in production, there isn't much security benefit to externalizing it. More importantly, introducing Spring property injection into this Mockito-only test would add unnecessary complexity.

Meanwhile, when you pointed this out, I also found that the default fallback string in JwtTokenUtil.java, "comixed-project", does not meet the minimum length requirement for the new algorithm, which uses Keys.hmacShaKeyFor(...) for token generation. It requires a key of at least 32 bytes.

So, I am updating the default key in both the @Value annotation and comixed-core.properties:

@Value("${comixed.auth.jwt-signing-key:comixed-project-secure-test-key-32}")
String signingKey;
comixed.auth.jwt-signing-key=comixed-project-secure-test-key-32

This also makes the default key consistent with the test key.

2. I will add my name to CONTRIBUTORS.md.



So, my changes for this are:

Update the default signingKey to meet the minimum 32-byte requirement and make it consistent with the test key.
Update CONTRIBUTORS.md.

mcpierce: Thanks for the clarification! I didn't realize I was looking at the testing key.

mcpierce:
@zoomgy Please squash the two commits down to a single one with the original commit message, and I'll merge it.

comixed-app/src/main/resources/comixed-core.properties
```diff
@@ -8,7 +8,7 @@ server.port=7171
 server.servlet.context-path=/
 spring.h2.console.enabled=false
 spring.h2.console.path=/dbconsole
-comixed.auth.jwt-signing-key=comixed-project
+comixed.auth.jwt-signing-key=comixed-project-secure-test-key-32
```
mcpierce: One quick change: please remove "-test" from the default string.

zoomgy: Okay what you want me to give as the default key ?

mcpierce: Yeah. How about moving it to the application.properties file so that admins can change it if they want to use a different value? What do you think?

(eta)

Let's make the string "comixed-project-secure-key-32" and put it into the application.properties file so it can be changed easily.

zoomgy: I think moving it to application.properties makes sense if the intention is for administrators to be able to configure the JWT signing key. The property is currently already externalised in comixed-core.properties, so moving it to application.properties would mainly make it part of the application-level configuration. I don't see any concerns with making this change.

mcpierce: For clarifification: **comixed-core.properties** isn't an external file, per se. It's shipped in the JAR file and is intended to hold properties that should not be overridden. When I mention externalizing, I mean putting it where the admin can change it.

zoomgy: "comixed-project-secure-key-32" is less than 32 bytes and cannot work with the latest signing algo, Keys.hmacShaKeyFor(...), we need at least 32 bytes for HS256.

My suggestions 
"comixed-jwt-signing-key-2026-32byte"

zoomgy: > For clarifification: **comixed-core.properties** isn't an external file, per se. It's shipped in the JAR file and is intended to hold properties that should not be overridden. When I mention externalizing, I mean putting it where the admin can change it.

Ohh! okay thanks for clarification, I will keep this in mind.

zoomgy: > "comixed-project-secure-key-32" is less than 32 bytes and cannot work with the latest signing algo, Keys.hmacShaKeyFor(...), we need at least 32 bytes for HS256.
> 
> My suggestions "comixed-jwt-signing-key-2026-32byte"

Any Suggestions here ?

mcpierce: How about "comixed-secured-jwt-signing-key!", which is exactly 32-bytes and the text calls out its purpose? :smile:

zoomgy: I have done the changes and squashed the commit into a single commit.

mcpierce (approved):

mcpierce:
Excellent! Thank you, @zoomgy !

zoomgy:
> Excellent! Thank you, @zoomgy !

I'm still new to open-source contribution and would really like to upskill and contribute more. Since you maintain several projects, if there are any issues across ComiXed or your other projects that you think would be good opportunities for a beginner to take on, I'd be very grateful for your suggestions.

mcpierce:
@zoomgy My advice would be to run a server, look at the existing backlog of tickets, and submit PRs for anything that catches your eye that feels like a challenge.

